PR #3, opened 2017-02-07
fix compile errors with visual c++ 2015

csarn:
With the following changes, I was able to compile binder on windows 7 x64 with VC2015:
add <cctype> include
replace and/or/not keywords by &&/||/!
replace uint by unsigned int

### DIFF
--- a/binder/binder.cpp
+++ b/binder/binder.cpp
@@ -123,12 +123,12 @@ public:
 
 	virtual bool VisitFunctionDecl(FunctionDecl *F)
 	{
-		if( F->isCXXInstanceMember() or isa<CXXMethodDecl>(F) ) return true;
+		if( F->isCXXInstanceMember() || isa<CXXMethodDecl>(F) ) return true;
 
 		if( binder::is_bindable(F) ) {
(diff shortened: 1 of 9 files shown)


lyskov:
Thank you for the PR @csarn ! Could you please merge latest master into your branch and resolve merge conflicts so i can merge this?

lyskov:
Also, regarding the and/or replacement: i do not think this is needed, it is looks looks like your project might not be using C++11 setting and thats why usage of and/or trigger the errors. (see http://en.cppreference.com/w/cpp/language/operator_alternative). Note that you absolutely do need C++11 support or generated bindings will be impossible to compile.

csarn:
Thanks for your response! From what I gathered, Visual C++ 2015 has no switch for enabling C++11, as it is by default compatible with it. As it turns out, there is a compiler switch "/Za" for "disabling compiler extensions". The non-support of those alternative operator keywords seems to be a compiler "extension", so disabling those extensions in turn enables those keywords again. https://msdn.microsoft.com/en-us/library/34h23df8.aspx I will try with that switch and see if it works.

PR #4, opened 2017-02-11
enum include should also reference definition headers

thisiscam:
The reasoning is that we are using the alternative values of the values in the generated code, not just the declaration


lyskov:
Thank you for the PR @thisiscam ! This patch sounds good to me but could you please elaborate why this is needed? Is that your enum have forward declaration header or something?

thisiscam:
Yes it was a forward decl. Not sure if it applies to other things class though

PR #32, opened 2017-10-03
Add support for per-class includes.

asford:
Add support for per-class includes to binding generator.  Adding
"include_for_class" option to binder, which triggers inclusion of a
header on a class-specific basis. This features supports the definition
of per-class custom binding code in a standalone header, which may then
be referenced via the +binder option.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -408,6 +408,16 @@ void ClassBinder::request_bindings_and_skipping(Config const &config)
 /// extract include needed for this generator and add it to includes vector
 void ClassBinder::add_relevant_includes(IncludeSet &includes) const
 {
+	string const qualified_name_without_template = standard_name( C->getQualifiedNameAsString() );
+	std::map<string, std::vector<string>> const &per_class_includes= Config::get().per_class_includes();
+
+	if( per_class_includes.count(qualified_name_without_template) ) {
+		for(auto i : per_class_includes.at(qualified_name_without_template))
+		{
+			includes.add_include(i);
+		}
+	}
+
 	for(auto & m : prefix_includes ) binder::add_relevant_includes(m, includes, 0);
 	binder::add_relevant_includes(C, includes, 0);
 
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -10,6 +10,7 @@
 /// @brief  Binding generation for C++ struct and class objects
 /// @author Sergey Lyskov
 
+#include <binder.hpp>
 #include <class.hpp>
 #include <function.hpp>
 #include <enum.hpp>
@@ -414,7 +415,8 @@ void ClassBinder::add_relevant_includes(IncludeSet &includes) const
 	auto pci = per_class_includes.find(qualified_name_without_template);
 	if(pci != per_class_includes.end()) {
 		for(auto const & i : pci->second) {
-			includes.add_include(i);
+			if(O_annotate_includes) includes.add_include(i + " // +include_for_class");
+			else includes.add_include(i);
 		}
 	}
 
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -410,10 +410,10 @@ void ClassBinder::add_relevant_includes(IncludeSet &includes) const
 {
 	string const qualified_name_without_template = standard_name( C->getQualifiedNameAsString() );
 	std::map<string, std::vector<string>> const &per_class_includes= Config::get().per_class_includes();
-
-	if( per_class_includes.count(qualified_name_without_template) ) {
-		for(auto i : per_class_includes.at(qualified_name_without_template))
-		{
+	
+	auto pci = per_class_includes.find(qualified_name_without_template);
+	if(pci != per_class_includes.end()) {
+		for(auto const & i : pci->second) {
 			includes.add_include(i);
 		}
 	}
--- a/documentation/config.rst
+++ b/documentation/config.rst
@@ -87,6 +87,14 @@ Config file directives:
   -include <boost/format/internals.hpp>
   +include <python/PyRosetta/binder/stl_binders.hpp>
 
+* ``include_for_class``, directive to control C++ include directives on a per-class basis. Force Binder to add particular include
+  into generated source files when a given target class is present. This allows the inclusion of custom binding code, which may
+  then be referenced with either ``+binder`` or ``+add_on_binder`` directives.
+
+.. code-block:: bash
+
+  +include_for_class example::class <example/class_binding.hpp>
+
 
 * ``binder``, specify custom binding function for particular concreate or template class. In the example below all
   specializations of tempalte std::vector will be handled by ``binder::vector_binder`` function. For template classes binder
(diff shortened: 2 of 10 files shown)


source/class.cpp
```diff
@@ -408,6 +408,16 @@ void ClassBinder::request_bindings_and_skipping(Config const &config)
 /// extract include needed for this generator and add it to includes vector
 void ClassBinder::add_relevant_includes(IncludeSet &includes) const
 {
+	string const qualified_name_without_template = standard_name( C->getQualifiedNameAsString() );
+	std::map<string, std::vector<string>> const &per_class_includes= Config::get().per_class_includes();
+
+	if( per_class_includes.count(qualified_name_without_template) ) {
```
lyskov: Right now there is two lookup for `qualified_name_without_template` which is not good. Could you please refactor this and use `map.find` and then examine and use it result. You can find example of this technique here: https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L358

source/class.cpp
```diff
@@ -408,6 +408,16 @@ void ClassBinder::request_bindings_and_skipping(Config const &config)
...
+		for(auto i : per_class_includes.at(qualified_name_without_template))
```
lyskov: Please use `auto const &` to avoid creating extra copy

source/class.cpp
```diff
@@ -408,6 +408,16 @@ void ClassBinder::request_bindings_and_skipping(Config const &config)
...
+		{
```
lyskov: Please move bracket to previous line so formatting is consistent

asford:
I'm trying to add test coverage for this feature, but noticing that `self-test.py` does not function on my macos machine. Is the test script intended to support macos?

asford:
I'm seeing failures in `build-and-run-tests.py` on an ubuntu machine, is this expected?

````
T10.inheritance.hpp Running test...
/home/fordas/workspace/binder_update_1710/binder/build/llvm-4.0/build_release_40.linux.digs.release/bin/binder --bind "" --root-module T10_inheritance --prefix /home/fordas/workspace/binder_update_1710/binder/test/build/ --single-file --annotate-includes  /home/fordas/workspace/binder_update_1710/binder/test/build//T10.inheritance.hpp.include -- -x c++ -std=c++11 -I /home/fordas/workspace/binder_update_1710/binder/test -I /home/fordas/workspace/binder_update_1710/binder/test/..
Generate bindings, pass 1...
Generate bindings, pass 2...
Sorting Binders...
Sorting Binders... Done.
Writing code...
Writing code... Done.

T10.inheritance.hpp Compiling binding results...
cd /home/fordas/workspace/binder_update_1710/binder/test/build/ && clang++ -O3 -shared -std=c++11 -I /home/fordas/workspace/binder_update_1710/binder/build/pybind11/include -I/usr/include/python2.7 -I./.. -I./../.. -I./../../source T10_inheritance.cpp -o T10_inheritance.so -fPIC

T10.inheritance.hpp Testing imports...
cd /home/fordas/workspace/binder_update_1710/binder/test/build/ && python -c 'import T10_inheritance'

Comparing results for test T10.inheritance.hpp...
diff /home/fordas/workspace/binder_update_1710/binder/test/T10.inheritance.ref /home/fordas/workspace/binder_update_1710/binder/test/build//T10_inheritance.cpp
6,7d5
< #include <T10.inheritance.hpp> // B::B
< #include <T10.inheritance.hpp> // B::operator=
245,247d242
< 		cl.def(pybind11::init<const class B &>(), pybind11::arg(""));
<
< 		cl.def("assign", (class B & (B::*)(const class B &)) &B::operator=, "C++: B::operator=(const class B &) --> class B &", pybind11::return_value_policy::automatic, pybind11::arg(""));


Encounter error while executing: diff /home/fordas/workspace/binder_update_1710/binder/test/T10.inheritance.ref /home/fordas/workspace/binder_update_1710/binder/test/build//T10_inheritance.cpp

Encounter error while executing: diff /home/fordas/workspace/binder_update_1710/binder/test/T10.inheritance.ref /home/fordas/workspace/binder_update_1710/binder/test/build//T10_inheritance.cpp
6,7d5
< #include <T10.inheritance.hpp> // B::B
< #include <T10.inheritance.hpp> // B::operator=
245,247d242
< 		cl.def(pybind11::init<const class B &>(), pybind11::arg(""));
<
< 		cl.def("assign", (class B & (B::*)(const class B &)) &B::operator=, "C++: B::operator=(const class B &) --> class B &", pybind11::return_value_policy::automatic, pybind11::arg(""));< #include <T10.inheritance.hpp> // B::B
< #include <T10.inheritance.hpp> // B::operator=
245,247d242
< 		cl.def(pybind11::init<const class B &>(), pybind11::arg(""));
<
< 		cl.def("assign", (class B & (B::*)(const class B &)) &B::operator=, "C++: B::operator=(const class B &) --> class B &", pybind11::return_value_policy::automatic, pybind11::arg(""));
````

**Comment by lyskov** [comment]
I have not tested self-test on Mac OS so it probably expected. 
Re Ubuntu self test run: This look like aftermath of LLVM upgrade and probably could be safe to ignore.

**Comment by asford** [comment]
Should I update the test reference results to reflect these changes?

File: source/class.cpp (hunk 1 above)
Comment (asford) [reply source/class.cpp]: Happy to use the idiom here, but I'd argue that this is less readable and really no more performant that .count/.at. Repeated lookups in small maps are exceptionally fast, as the data structure is sitting in cache for the second lookup.

File: source/class.cpp [hunk 4]
```diff
@@ -408,6 +408,16 @@ void ClassBinder::request_bindings_and_skipping(Config const &config)
 /// extract include needed for this generator and add it to includes vector
 void ClassBinder::add_relevant_includes(IncludeSet &includes) const
 {
+	string const qualified_name_without_template = standard_name( C->getQualifiedNameAsString() );
+	std::map<string, std::vector<string>> const &per_class_includes= Config::get().per_class_includes();
+	
+	auto pci = per_class_includes.find(qualified_name_without_template);
+	if(pci != per_class_includes.end()) {
+		for(auto const & i : pci->second) {
+			includes.add_include(i);
```
lyskov: Looks good Alex, thank you for refactoring this!

Small note: could you please add `if(O_annotate_includes)` and when annotation is requested add comment `// +include_for_class` for these includes so it is obvious when they come from? (this will be really helpful for any include-debugging later)

https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L188

lyskov:
> Should I update the test reference results to reflect these changes?
 If it only contain diff set from above then yes, but it is more changes then please skip this and i will do it later - i want to make sure that there is no unexpected changes

source/class.cpp
asford: Good idea, updated along with expected test results.

asford:
There are multiple reference result changes in the current `master`:

````
test/T10.inheritance.ref
test/T11.override.ref
test/T40.stl.ref
test/T42.stl.names.ref
````

I'm holding off on modifying these reference results, though they look benign to me. Filed under #33.

lyskov:
Looks good to me - merging!

PR #44, opened 2018-04-13
Build binder from llvm source download.

asford:
Update build.py to build from llvm source download, rather than git checkout.
This change significantly reduces the required disc space and checkout time for
a binder build by eliminating the full git clone of the llvm and clang.

Updates build invocation to *only* build binder, rather than the full
collection of clang and llvm tools, reducing the total build time.

### DIFF
--- a/build.py
+++ b/build.py
@@ -86,17 +86,26 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True, 
         Return absolute path to executable on success and terminate with error on failure
     '''
     release = 'release_40'
-    prefix += '/llvm-4.0'
-
-    git_checkout = '( git checkout {0} && git reset --hard {0} )'.format(release) if clean else 'git checkout {}'.format(release)
-
     if not os.path.isdir(prefix): os.makedirs(prefix)
 
-    if not os.path.isdir(prefix+'/.git'): execute('Clonning llvm...', 'cd {} && git clone https://github.com/llvm-mirror/llvm.git .'.format(prefix) )
-    execute('Checking out LLVM revision: {}...'.format(release), 'cd {prefix} && ( {git_checkout} || ( git fetch && {git_checkout} ) )'.format(prefix=prefix, git_checkout=git_checkout) )
+    llvm_version='4.0.0'
+    prefix += '/llvm-' + llvm_version
+    clang_path = "{prefix}/tools/clang".format(**locals())
 
-    if not os.path.isdir(prefix+'/tools/clang'): execute('Clonning clang...', 'cd {}/tools && git clone https://github.com/llvm-mirror/clang.git clang'.format(prefix) )
-    execute('Checking out Clang revision: {}...'.format(release), 'cd {prefix}/tools/clang && ( {git_checkout} || ( git fetch && {git_checkout} ) )'.format(prefix=prefix, git_checkout=git_checkout) )
+
+    if not os.path.isdir(prefix):
+        execute(
+            "Download llvm source.",
+            "curl http://releases.llvm.org/{llvm_version}/llvm-{llvm_version}.src.tar.xz | tar -Jxo &&"
+            "mv llvm-{llvm_version}.src {prefix}"
+            .format(llvm_version=llvm_version, prefix=prefix))
+
+    if not os.path.isdir(clang_path):
+        execute(
+            "Download clang source.",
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -86,17 +86,26 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True,
         Return absolute path to executable on success and terminate with error on failure
     '''
     release = 'release_40'
-    prefix += '/llvm-4.0'
```
lyskov: Should `release = 'release_40'` line above be removed now?

build.py
```diff
@@ -86,17 +86,26 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True,
...
+    clang_path = "{prefix}/tools/clang".format(**locals())
 
-    if not os.path.isdir(prefix): os.makedirs(prefix)
 
-    if not os.path.isdir(prefix+'/.git'): execute('Clonning llvm...', 'cd {} && git clone https://github.com/llvm-mirror/llvm.git .'.format(prefix) )
-    execute('Checking out LLVM revision: {}...'.format(release), 'cd {prefix} && ( {git_checkout} || ( git fetch && {git_checkout} ) )'.format(prefix=prefix, git_checkout=git_checkout) )
+    if not os.path.isdir(prefix):
+        execute(
+            "Download llvm source.",
+            "curl http://releases.llvm.org/{llvm_version}/llvm-{llvm_version}.src.tar.xz | tar -Jxo &&"
```
lyskov: optional, but lets keep things pretty: this line missing space after last `&&`

build.py
```diff
@@ -86,17 +86,26 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True,
...
+            "mv llvm-{llvm_version}.src {prefix}"
+            .format(llvm_version=llvm_version, prefix=prefix))
 
-    if not os.path.isdir(prefix+'/tools/clang'): execute('Clonning clang...', 'cd {}/tools && git clone https://github.com/llvm-mirror/clang.git clang'.format(prefix) )
-    execute('Checking out Clang revision: {}...'.format(release), 'cd {prefix}/tools/clang && ( {git_checkout} || ( git fetch && {git_checkout} ) )'.format(prefix=prefix, git_checkout=git_checkout) )
+    if not os.path.isdir(clang_path):
+        execute(
+            "Download clang source.",
+            "curl http://releases.llvm.org/{llvm_version}/cfe-{llvm_version}.src.tar.xz | tar -Jxo &&"
```
lyskov: Same here: missing space after last `&&`

build.py
```diff
@@ -160,6 +169,7 @@ def main(args):
     parser.add_argument('--pybind11', default='', help='Path to pybind11 source tree')
     parser.add_argument('--annotate-includes', action="store_true", help='Annotate includes in generated source files')
     parser.add_argument('--trace', action="store_true", help='Binder will add trace output to to generated source files')
+    parser.add_argument('--standalone-install', default="", help='Install standalone binder binary at given installation prefix.')
```
lyskov: This line need to be removed? (ie unused option)

build.py
```diff
@@ -86,17 +86,26 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True,
...
-    git_checkout = '( git checkout {0} && git reset --hard {0} )'.format(release) if clean else 'git checkout {}'.format(release)
+    llvm_version='4.0.0'
+    prefix += '/llvm-' + llvm_version
+    clang_path = "{prefix}/tools/clang".format(**locals())
 
-    if not os.path.isdir(prefix): os.makedirs(prefix)
 
-    if not os.path.isdir(prefix+'/.git'): execute('Clonning llvm...', 'cd {} && git clone https://github.com/llvm-mirror/llvm.git .'.format(prefix) )
-    execute('Checking out LLVM revision: {}...'.format(release), 'cd {prefix} && ( {git_checkout} || ( git fetch && {git_checkout} ) )'.format(prefix=prefix, git_checkout=git_checkout) )
+    if not os.path.isdir(prefix):
```
lyskov: Before i explicitly checked for `/.git` dir instead of just for prefix to exists. Reason for this is that if network connection hang and user terminate the script then dir will be created but it will be empty, so by checking that some content is exist inside the prefix we can avoid recovery by-hands in sub sequential runs. Therefor: could you please change line above to something like `if not os.path.isfile(prefix + '/CMakeLists.txt')`?

PR #50, opened 2018-09-09
Update README.md

DomHudson:
As the readme is the first thing many people see regarding the project, it makes sense to add a bit more information to it. This is just adding a small paragraph taken from the About page in the documentation.


lyskov:
Awesome, thank you @DomHudson !

PR #53, opened 2018-09-28
Example cmake script

kostyfisik:
1) Make it easier to use cmake example script with other projects. All paths and settings are in the single place (in the top of the script)
2) Unify usage of quotes (to visually stress the difference from C\C++ I selected single quotes, however, it is not a problem to put all of them to be double quotes).  
3) Add other often used file extensions widely used in C++ projects.
4) In many projects they use separate include and source dir. I am not sure whether line "for include_dir in [binder_source, .." is OK now for Binder, as it includes both this_project_source and this_project_include (which are the same for example_structure project).
5) Added missing config file which is also needed for make_binding_via_bash_and_stl.sh

### DIFF
--- a/examples/example_struct/config.cfg
+++ b/examples/example_struct/config.cfg
@@ -1,0 +1,1 @@
++include <pybind11/stl.h>
--- a/examples/example_struct/make_bindings_via_cmake.py
+++ b/examples/example_struct/make_bindings_via_cmake.py
@@ -8,83 +8,97 @@ import subprocess
 from distutils.sysconfig import get_python_inc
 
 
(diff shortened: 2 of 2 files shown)


lyskov:
Bringing in @danpf who wrote the original script. Daniel could you please take a look at this PR?

danpf:
@kostyfisik thank you for your contribution! This looks good to me! 
RE point 4:
I think that this is fine. This is just meant to give people an idea of how they might get started with making bindings for a project.  Once I understood the basics I found it easy to translate into other projects, and include errors are usually self explanatory.  If you have the urge though, feel free to make a more complex example!

kostyfisik:
> If you have the urge though, feel free to make a more complex example!

Ok, I will provide such an example as soon as the PR is merged\squashed.

PR #54, opened 2018-10-08
Danpf/remove trace example

danpf:
Don't confuse people by having -trace statements in the cmake example.

I meant to do this originally but I had forgotten.


danpf:
@lyskov sorry, could you approve again? I dismissed it by accident when I merged with master

PR #59, opened 2019-01-28
Few tweaks to support building with MSVC

jquesnelle:
Binder can be built with MSVC (I used VS 2017) with only a few minor tweaks. They are:

- Alias `uint` to `unsigned int`
  - I'm not sure how in the include chain `uint` gets defined on Linux or Clang/gcc, but for MSVC it's undefined
- Include `<ciso646>`
  - MSVC by default does not support the alternative operator representations (e.g. `or`, `and`). However, they can be added in by including `<ciso646>`

### DIFF
--- a/source/config.hpp
+++ b/source/config.hpp
@@ -18,6 +18,11 @@
 #include <vector>
 #include <map>
 
+#ifdef _MSC_VER
+#include <ciso646>
+using uint = unsigned int;
+#endif
+
 namespace binder {
 
 class Config


lyskov:
Thank you for putting this together @jquesnelle ! Looks good to me, i will merge this in a moment. 

Btw MSVC does support `and` and `or` C++11 keywords but it require additional command line flag to be specified, please see https://docs.microsoft.com/en-us/cpp/build/reference/za-ze-disable-language-extensions?view=vs-2017 for details.

lyskov:
Thank you for putting this together @jquesnelle ! Looks good to me, i will merge this in a moment. 

Btw MSVC does support `and` and `or` C++11 keywords but it require additional command line flag to be specified, please see https://docs.microsoft.com/en-us/cpp/build/reference/za-ze-disable-language-extensions?view=vs-2017 for details.

PR #60, opened 2019-01-28
Support Clang 7

jquesnelle:
I'm prototyping using binder with Visual Studio 2017, which means that it needs to parse the MSVC STL. This *mostly* works, but not completely. For VS 2017, Clang needs the `-fms-compatibility-version=20` flag, which is only supported in Clang 7 (Clang 4 only has up to `-fms-compatibility-version=19`). Turns out there aren't too many changes needed to make binder build with LLVM/Clang 7! They are:

- Remove `using fmt::format` and change calls to `format` to `fmt::format`
  - At some point Clang introduced their own `format` function, which conflicts with the one binder builds in. Using the namespaced version resolves to the correct function
- Disambiguate between `llvm::PointeeType`, `clang::PointeeType`, `llvm::Type`, and `clang::Type`
  - Along the some vein, LLVM added in a `PointeeType` and `Type`; and since we're `using namespace llvm` and `using namespace clang` in `type.cpp`, there's a conflict
- Update `CMakeLists.txt` to use private target libraries for Clang 6+
  - The Clang build system was updated with Clang 6, and targets need to mark their libraries as private. It's a little ugly, but marking the libraries as private in <= 5 actually fails, so I had to switch based on `LLVM_VERSION_MAJOR`

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -3,6 +3,7 @@
 #set(CMAKE_CXX_FLAGS "-fexceptions ${CMAKE_CXX_FLAGS}")
 
 set(LLVM_LINK_COMPONENTS support)
+set(LLVM_USED_LIBS clangTooling clangBasic clangAST)
 
 add_clang_executable(binder
   binder.cpp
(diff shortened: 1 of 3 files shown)


lyskov:
Closing as no longer relevant, - LLVM-7 support is now merged into master.

PR #65, opened 2019-04-13
a way to suppress errors for not-found included headers

yosoufe:
a way to suppress errors if the not-found included headers are not useful in generating bindings and you are sure that the bindings are still correct and you want to suppress ugly fatal errors then you can use `--suppress-errors=true` as argument to get rid of them.


lyskov:
Also, @yosoufe could you please add one-line-description of this new options into documentation? I think putting it into this page might be best: https://cppbinder.readthedocs.io/en/latest/config.html

yosoufe:
@lyskov Thanks. Done.

PR #67, opened 2019-04-16
a way to suppress errors if the not-found included headers

yosoufe:
are not useful in generating bindings and you are sure that
the bindings are still correct and you want to suppress ugly
fatal errors then you can use --suppress-errors=true as argument
to get rid of them.

Sorry. This is the same as https://github.com/RosettaCommons/binder/pull/65
I had to rebranch stuff on my repo. So it closed the other PR.


lyskov:
- thank you for implementing this @yosoufe! I will merge this into `master` as soon as i test this on Rosetta test suite.

PR #69, opened 2019-04-18
Use variable names as well for overloaded constructors

yosoufe:
Use variable names for overloaded constructors instead of a1, a2 ...
And more importantly add `pybind11::arg` to get proper helpers inside the python for overloaded constructors with arguments with default values.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -882,6 +882,7 @@ struct ConstructorBindingInfo
 //char const * constructor_if_template = "\tcl.def(\"__init__\", [cl_type](pybind11::handle self_{0}) {{ if (self_.get_type() == cl_type) new (self_.cast<{2} *>()) {2}({1}); else new (self_.cast<{3} *>()) {3}({1}); }}, \"doc\");";
 
 char const * constructor_template =                 "\tcl.def( pybind11::init( []({0}){{ return new {2}({1}); }} ), \"doc\");";
+char const * constructor_template_with_pyArg =      "\tcl.def( pybind11::init( []({0}){{ return new {2}({1}); }} ), \"doc\" {3});";
 char const * constructor_with_trampoline_template = "\tcl.def( pybind11::init( []({0}){{ return new {2}({1}); }}, []({0}){{ return new {3}({1}); }} ), \"doc\");";
 //char const * constructor_lambda_template = "\tcl.def(pybind11::init( []({0}){{ return new {2}({1}); }}, []({0}){{ return new {3}({1}); }} ), \"doc\");";
 
@@ -910,6 +911,11 @@ string bind_constructor(ConstructorBindingInfo const &CBI, uint args_to_bind, bo
 
 		//string params = args_to_bind ? ", " + args.first : "";
 		string params = args_to_bind ? args.first : "";
+		
+		string argsHelper;
+		for(uint i=0; i<CBI.T->getNumParams()  and  i < args_to_bind; ++i) {
+			argsHelper += ", pybind11::arg(\"{}\")"_format( string( CBI.T->getParamDecl(i)->getName() ) );
+		}
 
 		// if( CBI.T->isVariadic() ) c = fmt::format(constructor_lambda_template, params, args.second, constructor_types.first, constructor_types.second);
 		// else if( constructor_types.first.size()  and  constructor_types.second.size() ) c = fmt::format(constructor_lambda_template, params, args.second, constructor_types.first, constructor_types.second);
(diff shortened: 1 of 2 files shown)


source/function.cpp
```diff
@@ -86,8 +86,8 @@ pair<string, string> function_arguments_for_lambda(clang::FunctionDecl const *re
 		QualType qt = record->getParamDecl(i)->getOriginalType().getCanonicalType();
 		r += qt.getAsString() + ' ';
 		if( !qt->isReferenceType()  and  !qt->isPointerType() ) r += !qt.isConstQualified() ? "const & " : "& ";
-		r += "a" + std::to_string(i);
-		a += "a" + std::to_string(i);
+		r += string( record->getParamDecl(i)->getName() );
+		a += string( record->getParamDecl(i)->getName() );
```
lyskov: Have you checked if this approach works when there is no name of argument? Like `MyClass(int, float)`

yosoufe: Nope. I forgot that. But why aren't you worried about the same thing but for the complete constructor with all arguments?

source/class.cpp
```diff
@@ -910,6 +911,11 @@ string bind_constructor(ConstructorBindingInfo const &CBI, uint args_to_bind, bo
 
 		//string params = args_to_bind ? ", " + args.first : "";
 		string params = args_to_bind ? args.first : "";
+		
+		string argsHelper;
+		for(uint i=0; i<CBI.T->getNumParams()  and  i < args_to_bind; ++i) {
+			argsHelper += ", pybind11::arg(\"{}\")"_format( string( CBI.T->getParamDecl(i)->getName() ) );
+		}
```
yosoufe: @lyskov  I reverted the change in `function.cpp`. my main purpose of this PR was to get documentation of the arguments for other constructors in Python environment that is mainly here. And does not need that change  in `function.cpp`.

source/function.cpp
yosoufe: Anyway I reverted this change. I do not need it to get the argument name in the documentation in python environment.

lyskov: hm... i think there was some miscommunication here: i was not saying that this change is bad (i think it good!) - i was just asking if this change would break for cases when no name of arg was giving. Any chance you can re-introduce this change back? And if you can not check case of args-without-name - no worries, i will just check that locally. Thanks,

source/class.cpp
```diff
@@ -882,6 +882,7 @@ struct ConstructorBindingInfo
 //char const * constructor_if_template = "\tcl.def(\"__init__\", [cl_type](pybind11::handle self_{0}) {{ if (self_.get_type() == cl_type) new (self_.cast<{2} *>()) {2}({1}); else new (self_.cast<{3} *>()) {3}({1}); }}, \"doc\");";
 
 char const * constructor_template =                 "\tcl.def( pybind11::init( []({0}){{ return new {2}({1}); }} ), \"doc\");";
+char const * constructor_template_with_pyArg =      "\tcl.def( pybind11::init( []({0}){{ return new {2}({1}); }} ), \"doc\" {3});";
```
lyskov: sorry, a bit pedantic but i have to keep code tidy: could you please name variable in the same style as rest of the code (ie snake-case), like: `constructor_template_with_py_arg`

source/class.cpp
```diff
@@ -918,6 +924,7 @@ string bind_constructor(ConstructorBindingInfo const &CBI, uint args_to_bind, bo
 
 		if( CBI.C->isAbstract() ) c = fmt::format(constructor_template, params, args.second, CBI.trampoline_qualified_name);
 		else if( CBI.trampoline ) c = fmt::format(constructor_with_trampoline_template, params, args.second, CBI.class_qualified_name, CBI.trampoline_qualified_name);
+		else if( args_to_bind )c = fmt::format(constructor_template_with_pyArg, params, args.second, CBI.class_qualified_name, argsHelper);
 		else c = fmt::format(constructor_template, params, args.second, CBI.class_qualified_name);
```
lyskov: is this case became unreachable? And if so then `constructor_template` will need to be commented out.

yosoufe: I do not think it is going to be unreachable. For a case that all constructors have default values, the last case is reachable.

yosoufe: Done. Thank you.

source/class.cpp
yosoufe: Done

source/function.cpp
yosoufe: Thank you.

PR #58, opened 2018-12-17
Danpf/uninitialized copy fix

danpf:
Added docs for debugging + a fix for unitialized_copy bindings.

PR #73, opened 2019-05-16
Danpf/uninitialized copy fix (binder fixes only)

danpf:
writing good documentation is hard and takes a lot of time, lets just push the core binder changes through in this PR while I wait until i have a spare moment or two to revisit the compiling documentation that I started on https://github.com/RosettaCommons/binder/pull/58/files

After looking through it, I'm not sure it's professional or thorough for the public documentation... (yet)

@kostyfisik track this, it should get merged quickly

### DIFF
--- a/examples/example_struct/make_bindings_via_cmake.py
+++ b/examples/example_struct/make_bindings_via_cmake.py
@@ -14,25 +14,30 @@ bindings_dir = 'cmake_bindings'
 binder_source = f'{os.getcwd()}/../../source'
 pybind_source = f'{os.getcwd()}/../../build/pybind11/include'
 use_pybind_stl = True
-#use_pybind_stl = False
+# use_pybind_stl = False
 this_project_source = f'{os.getcwd()}/include'
 this_project_include = this_project_source
 this_project_namespace_to_bind = 'testers'
 python_module_name = 'test_struct'
 
+
 def make_all_includes():
     all_includes = []
     all_include_filename = 'all_cmake_includes.hpp'
-    for filename in (glob.glob(f'{this_project_source}/**/*.hpp', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.cpp', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.h', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.cc', recursive=True)+
+    for filename in (glob.glob(f'{this_project_source}/**/*.hpp', recursive=True) +
+                     glob.glob(f'{this_project_source}/**/*.cpp', recursive=True) +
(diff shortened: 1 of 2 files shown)


examples/example_struct/make_bindings_via_cmake.py
```diff
@@ -14,25 +14,30 @@
...
     all_includes = []
     all_include_filename = 'all_cmake_includes.hpp'
-    for filename in (glob.glob(f'{this_project_source}/**/*.hpp', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.cpp', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.h', recursive=True)+
-                     glob.glob(f'{this_project_source}/**/*.cc', recursive=True)+
+    for filename in (glob.glob(f'{this_project_source}/**/*.hpp', recursive=True) +
+                     glob.glob(f'{this_project_source}/**/*.cpp', recursive=True) +
+                     glob.glob(f'{this_project_source}/**/*.h', recursive=True) +
+                     glob.glob(f'{this_project_source}/**/*.cc', recursive=True) +
```
lyskov: This seems quite repetitive and repetitive... So how about something like this:
```Python
for ext in 'hpp cpp h cc c'.split():
   for filename in (glob.glob(f'{this_project_source}/**/*.' + ext, recursive=True):
```

examples/example_struct/make_bindings_via_cmake.py
```diff
@@ -48,8 +53,8 @@ def make_bindings_code(all_includes_fn):
                + ('--config config.cfg ' if use_pybind_stl else '') +
                f' {all_includes_fn} -- -std=c++11 '
                f'-I{this_project_include} -DNDEBUG -v').split()
-    print(' '.join(command))
-    ret = subprocess.call(command)
+    print('BINDER COMMAND:', ' '.join(command))
+    subprocess.call(command)
```
lyskov: Here and in other places of this script: it probably a good idea to replace `subprocess.call` with `subprocess.check_call` so errors during call terminate the script (which is what we want most of the time). (see https://docs.python.org/3/library/subprocess.html#subprocess.check_call for details)

examples/example_struct/make_bindings_via_cmake.py
danpf: I agree, but I think it is better to keep it the way it is for verbosity in this example.

examples/example_struct/make_bindings_via_cmake.py
danpf:
```suggestion
    subprocess.check_call(command)
```

danpf:
I switched to check_call!, sorry, need a re-review! thank you @lyskov

PR #82, opened 2019-08-27
Fix binder_for_namespace docs

simleo:
Fixes the config example to refer to `binder_for_namespace` instead of `add_on_binder_for_namespace`.


lyskov:
Thank you @simleo !

PR #88, opened 2020-01-03
Added a simple standalone CMakeLists.txt that can be used with  installed clang/llvm

andriish:
Hi all, 
these changes  can be useful for cases when clang/llvm are already installed.
Work fine at lest with CentOS8 and with CentOS7+ llvm-toolset-7. Should not interfere with the default way of compilation.
Andrii


lyskov:
Thank you for the patch @andriish! Could you please add docs for this case to outline how exactly Binder could be built with LLVM installed system-wide?

andriish:
Dear @lyskov ,

1) thank you for considering the patch.

2) The process to build binder itself in the simplest case  is trivial and consist only of 
``
cmake CMakeLists.txt  -DCMAKE_INSTALL_PREFIX:PATH=/home/user/whereiwanttohaveit/
make
make install
``
The location of LLVM and CLANG can be set separately.


The preparation of dependencies, however, is less trivial, so I will describe the process in detail at least for some systems.
a) Everything should work out of the box for the Fedora31 after running as root
``
yum install clang clang-devel llvm-devel llvm-static clang-libs libcxx libcxx-devel
``

b) For the CentOS8/RHEL8  the dependencies are the same, but libcxx and  libcxx-devel are not available yet from standard repos ( should be available soon?)
so  one has to recompile it.
``
rpm -i http://ftp.tu-chemnitz.de/pub/linux/fedora/linux/releases/31/Everything/source/tree/Packages/l/libcxx-9.0.0-1.fc31.src.rpm
wget https://src.fedoraproject.org/rpms/libcxx/raw/f31/f/libcxx.spec
``
edit  libcxx.spec   to set
``
%global bootstrap 1
``
Then 
``
rpmbuild -bb libcxx.spec
``
And as root

``
rpm -Uvh /home/user/rpmbuild/RPMS/x86_64/libcxx-devel-9.0.0-1.el8.x86_64.rpm /home/user/rpmbuild/RPMS/x86_64/libcxx-9.0.0-1.el8.x86_64.rpm 
``
c) For the CentOS7 there are two options to build. The first one is to use llvm-toolset-7.0 
https://www.softwarecollections.org/en/scls/rhscl/llvm-toolset-7.0/
(as root)
``
yum install llvm-toolset-7.0*  libcxx libcxx-devel
``
and use   as shell for compilation 
``
scl enable llvm-toolset-7 bash
``
The second option is to patch binder sources to make it work with the default clang-3.4.2 and llvm-3.4.2. I can provide patches for this case.


Best regards,
Andrii

andriish:
Of course for the full scale installation one would need to take care about the documentation and some other smaller things. But once the top level CMkeLIsts.txt will be there it will be much easier to do.

source/CMakeLists.txt
```diff
@@ -39,4 +39,9 @@ target_link_libraries(binder
   clangTooling
   clangBasic
   clangASTMatchers
+  clangAST
+  clangFrontend
```
lyskov: Why `clangAST` and `clangFrontend` was added here? They are not needed for Binder build.

CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.4.3)
```
lyskov: Please add comment here outlining purpose of this file

source/CMakeLists.txt
andriish: Basically the linker advices to do so:

/usr/bin/ld: CMakeFiles/binder.dir/class.cpp.o: undefined reference to symbol '_ZNK5clang11DeclContext18isDependentContextEv'
/opt/rh/llvm-toolset-7/root/usr/lib64/libclangAST.so.5: error adding symbols: DSO missing from command line

/usr/bin/ld: CMakeFiles/binder.dir/binder.cpp.o: undefined reference to symbol '_ZN5clang22PCHContainerOperationsC1Ev'
/opt/rh/llvm-toolset-7/root/usr/lib64/libclangFrontend.so.5: error adding symbols: DSO missing from command line

It looks like it is related to the way the clang is compiled.

CMakeLists.txt
andriish: Done.

PR #96, opened 2020-02-21
Enable binder builds with multiple LLVM versions

andriish:
Dear authors,

The merge request includes 
- improvements in the CMakeLists for standalone builds
- corresponding documentation on installation
- some #if statements for specific LLVM versions in the codes

The builds were tested with 
CentOS7 llvm-3.4/gcc-4.8.5/cmake-3.4.16  and  llvm7.0/clang7.0/cmake-3.6.2 
Fedora rawhide (llvm-7.0 llvm-8.0 llvm-9.0 llvm-10.0)/gcc-10.0/cmake-3.16.3

In practice with this merge request binder could be built standalone with most widely used versions of LLVM/CLANG which actually enables a lot of opportunities not only for the user but for the developers as well, e.g. one could easily enable light weight and fast CI.

Best regards,
Andrii


andriish:
Hi @lyskov , to illustrate the just the changes I've added a CI configuration based on a standard Fedora, CentOS7 and CentOS8. The configuration is not very advanced, but it works.

Best regards,

Andrii

andriish:
Just implemented a working version of CI with tests that use cmake only.

andriish:
Hi @lyskov ,
thanks a lot for merging #95 .
I've syncronized this MR with master.
Of course this MR is a bit larger and requires some time for review, but I will be happy to answer any related questions.

Best regards,
Andrii

andriish:
Dear Sergey,

1) The workflow directory is used by github CI.
Basically in this setup, each time someone makes a commit the github CI starts dokers and executes
the entrypoint.sh script.  
The short results for my fork can be seen here: 
https://github.com/andriish/binder/actions
I've invited you to the fork so you can see the full log view.
For instance 
https://github.com/andriish/binder/runs/501360596?check_suite_focus=true
 So far I test 3 linux flavors for the branch CentOS7 and 5 for the branch Ubuntu18.

2) cmake directory is needed for cmake. The cmake build system has two ways to configure the needed packages. The first one is to rely on the  <SomeOtherSoftware>Config.cmake delivered by 
package SomeOtherSoftware and the second one is to rely on the Find<SomeOtherSoftware>.cmake script. The second option is the most common case. For many packages  cmake delivers the scripts Find<SomeOtherSoftware>.cmake ,  for instance  I have  in my system /usr/share/cmake/Modules/FindMPEG2.cmake and /usr/share/cmake/Modules/FindOpenSSL.cmake. However, for some less common packages these scripts don't exist so people provide them within their projects in the /cmake directory.  Also quite often there are some extra templates for instalation/deinstallation in the cmake directory.
Even shorter one can describe /cmake directory as an equivalent of /m4 directory for autotools.

Best regards,
Andrii

lyskov:
@andriish i have given this a bit more thoughts and currently i am not sure if supporting multiple LLVM versions is way to go. Could you please argue why this would be a worthy goal and what kind of benefits it will give to Binder users?

The way i see it is that the LLVM version and Binder version should be in lockstep the way CFE(Clang) and LLVM source are. We can see Binder to be a some sort of as special compiler, which output is C++ source code to generate Python bindings. Reason for such lockstep is because LLVM low level API is inherently unstable and carry no guarantee that certain functionality will work in future versions the say way as it work now. 

Major implication of this is that LLVM source that Binder uses should be considered **part of Binder package** and not a dependency. Because of that these LLVM sources do not need to be shared with rest of system and should be seen only as inner tool on which Binder relay to do it work. Such treatment is not unique to Binder,  if we look at how different Clang (and GCC) versions is installed on same system we will see that Clang package will use particular LLVM library and different version of such libraries is treated as *different packages*.  My current take is that Binder should take the same approach: lock LLVM version it uses and if using this LLVM as outside package dependency is not feasible then supply it as part of the Binder package.

Also, on practical technical terms: after looking at ifdef's in this PR diff in Binder sources i am skeptical that this could be a feasible path forward. While current discrepancy between LLVM API is somewhat minor such ifdef's already creating clutter in source which will be increase maintenance future costs. And future LLVM versions will most likely require more ifdef's and introduce further clutter.

re CI integration with GitHub: - i was planning to do this when i have "a minute" so PR which will add this functionality will be welcome. Can docker files and other files needed to set this up separate be separated  into separate PR and made use of current implementation of `build.py` and `self-test.py`? 

re testing on different OS's: you probably already noticed that current Binder test system is highly dependable on version of libstdc++ and complier used. So maybe full versions of unit tests should only be run on CentOS with Clang and on other system we can just check if generated sources is compilable?

andriish:
Dear @lyskov , 


>  The way i see it is that the LLVM version and Binder version should be in lockstep the way CFE(Clang) and LLVM source are. 

Exactly. In the end my suggestion is only that each version of binder 1.a   would target LLVM versions between b.c and e.d with b and e being very close.   This would allow to avoid bloated code.  It is said "All" in the MR title just because the codes now are quite lean and it took quite little time to compile with all easily reachable versions of LLVM. And, of course, right now there is only one version released.


>  Could you please argue why this would be a worthy goal and what kind of benefits it will give to Binder users?

If the installation would be easier, more people will be able to use it and in a better way. The typical usage I saw so far is a single installation per team outside of development environment. E.g. the developers do their development in a normal environment and then to generate bindings switch to some virtual machine or docker. That happens because the installation of LLVM could be painful and last for some time. 


>  We can see Binder to be a some sort of as special compiler, which output is C++ source code to generate Python bindings.

Exactly.

> Major implication of this is that LLVM source that Binder uses should be considered part of Binder package and not a dependency.

That is one way of doing that. One still could consider them as dependency. As far as I understand  another compiler project, flang,  is doing exactly that  https://github.com/flang-compiler/flang/wiki/Building-Flang  (modulo forks).
And that makes sense as flang (and binder)  projects are not part of LLVM.

>  Also, on practical technical terms: after looking at ifdef's in this PR diff in Binder sources i am skeptical that this could be a feasible path forward. While current discrepancy between LLVM API is somewhat minor such ifdef's already creating clutter in source which will be increase maintenance future costs. And future LLVM versions will most likely require more ifdef's and introduce further clutter.

Fully agree.  Supporting *all versions* of LLVM *simultaneously* is not an option in a long run, that is completely out of question.  But having a range of supported versions for each given binder version would look reasonable.  What is more important is that there would be versions of binder could be compiled on the most standard linux  distributions without the recompilation of LLVM. 

More practically: if the current "recommended"  LLVM version is 6, would that make sense to you to
Have #ifdefs that enable versions 3 - 6 or so,  make a release 1.0.1 or so and for each subsequent release explicitly state the supported version range for builds with external LLVM? 


> re CI integration with GitHub: - i was planning to do this when i have "a minute" so PR which will add this functionality will be welcome. Can docker files and other files needed to set this up separate be separated into separate PR and made use of current implementation of build.py and self-test.py?

Yes, that is possible. The only needed thing is to edit entrypoint.sh scripts.
However the compilation of LLVM in docker will take hours vs. minutes with external LLVM. The solution to that is to have  CI activated not on pull, but manually or according to some schedule, e.g. perform nightly builds.
Event better would be to have a mixed approach: builds with external LLVM activated on pull (as it is)  and nightly builds using build.py.
Something like this https://github.com/arduino/arduino-cli/blob/master/.github/workflows/nightly.yaml
should work fine. 
I will have a look and provide a setup for that,  but again  for this approach it would be nice to have an option to build with external LLVM in a wider range of versions.
Does it sound reasonable to you? 


Concerning the self-test.py: see below.


> re testing on different OS's: you probably already noticed that current Binder test system is highly dependable on version of libstdc++ and complier used. So maybe full versions of unit tests should only be run on CentOS with Clang and on other system we can just check if generated sources is compilable?

The compiler can be changed easily, of course. The python tests have an  option -BINDER_TEST_PYTHON_VERSIONS=...
If 0 is given, only the difference  between generated  sources and the references will be tested.
And if BINDER_USE_PYTHON_IN_TEST=OFF  is given, python imports will be disabled.
 Is it that what you mean?

self-test.py: Maybe that would make sense to use cmake for testing.  Potential benefit is that one could avoid hardcoded python version, compiler options, etc., to let cmake find all the locations and make the tests extremely portable.  The drawback is that the CMakeLists.txt will look a bit bloated.
What is your opinion?


Best regards,
Andrii

andriish:
Dear @lyskov ,

after some research I've found a project that could be a good example to follow from in the technical aspects:
https://github.com/CastXML/CastXML

That is a successor of gccxml developed by KitWare people, so best cmake practices included. From the point of view of portability and availability the project is doing well:
https://repology.org/project/castxml/versions
They also have two build modes as I've suggested:

https://github.com/CastXML
the normal build and the superbuild.

Best regards,
Andrii

lyskov:
@andriish let me think about this. I am a bit swamped with my main project though so it could take me some time to get to this.

andriish:
Hi @lyskov ,

no problem.  Meanwhile I have actually a bit more time so I could split the MR in smaller pieces and/or provide more explanations.

source/binder.cpp
```diff
@@ -13,6 +13,18 @@
 #include <binder.hpp>
 
 // Declares clang::SyntaxOnlyAction.
+#if  (LLVM_VERSION_MAJOR < 4)
+#include "clang/AST/ASTConsumer.h"
+#include "clang/AST/ASTContext.h"
+typedef clang::ASTConsumer* ASTConsumerPtr;
```
lyskov, line 19: when possible please use C++11 `using declarations they are much more readable

source/binder.cpp
```diff
@@ -13,6 +13,18 @@
...
+#endif
+#if (LLVM_VERSION_MAJOR >= 4)
+#include "clang/AST/ASTConsumer.h"
+#include "clang/AST/ASTContext.h"
+#include "clang/Frontend/FrontendActions.h"
+typedef std::unique_ptr<clang::ASTConsumer> ASTConsumerPtr;
```
lyskov, line 25: This seems wrong to me: name of the type imply a pointer but unique pointer was used instead. Also, i do not thing using typedef here is justified since type is only named in single place.

.github/workflows/centos7/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -l
+set -x
+uname -a 
+cat /etc/issue
+yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
lyskov, line 5: Why not use CentOS-7 default Clang and LLVM? They should be at version ~3.4? so test them seems to be logical thing to do (more on this below).

source/binder.cpp
```diff
@@ -13,6 +13,18 @@
 #include <binder.hpp>
 
 // Declares clang::SyntaxOnlyAction.
+#if  (LLVM_VERSION_MAJOR < 4)
```
lyskov, line 16: Please say a few words what range of version of LLVM do you propose Binder should support? We moved to use LLVM-4 almost immediately so i am not sure if supporting versions <4 makes much sense. Also, looks like most of if/def spaghetti code below is caused by supporting version below 4. So i think it would make more sense to just support only LLVM-4 and above.

source/binder.cpp
```diff
@@ -86,7 +99,9 @@ class ClassVisitor : public RecursiveASTVisitor<ClassVisitor>
 	virtual ~ClassVisitor() {}
 
 	virtual bool VisitEnumDecl(EnumDecl *record) {
+#if  (LLVM_VERSION_MAJOR >= 4)
```
lyskov, line 102: Why this ifdef needed?

source/binder.cpp
```diff
@@ -121,7 +136,12 @@ class BinderVisitor : public RecursiveASTVisitor<BinderVisitor>
 		if( O_config.size() ) config.read(O_config);
 		if( O_suppress_errors )	{
 			clang::DiagnosticsEngine& di = ci->getDiagnostics();
+#if  (LLVM_VERSION_MAJOR < 10)
 			di.setSuppressAllDiagnostics();
+#endif
+#if  (LLVM_VERSION_MAJOR >= 10)
```
lyskov, line 142: please use `#else` instead

source/binder.cpp
```diff
@@ -234,19 +254,32 @@ class BinderASTConsumer : public ASTConsumer
...
 class BinderFrontendAction : public ASTFrontendAction {
 public:
+#if  (LLVM_VERSION_MAJOR < 4)
+    virtual ASTConsumerPtr CreateASTConsumer(CompilerInstance &ci, StringRef file) {
+        return (ASTConsumerPtr)( new BinderASTConsumer(&ci) );
+#endif
+#if  (LLVM_VERSION_MAJOR >= 4)
     virtual std::unique_ptr<clang::ASTConsumer> CreateASTConsumer(CompilerInstance &ci, StringRef file) {
         return std::unique_ptr<ASTConsumer>( new BinderASTConsumer(&ci) );
+#endif
```
lyskov, line 264: Why this ifdef is needed? Both branches seems to do exactly the same...

test/diffbinder.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return EXIT_SUCCESS;
+}
+int main(int argc, char** argv)
+{
+    if (argc!=3) {
+        std::cout<<"Wrong number of arguments"<<std::endl;
+        return EXIT_FAILURE;
+    }
+    return COMPARE_ASCII_FILES(argv[1],argv[2]);
+}
```
lyskov, line 55: Could you please outline purpose of this file? Looks like it kind of re-implementation of `diff`, so why not use `diff` instead?

documentation/install.rst
```diff
@@ -121,5 +123,17 @@ To build ``binder`` run
...
+``cmake CMakeLists.txt   -DLLVM_DIR=/usr/lib64/llvm8.0/lib/cmake/llvm -DClang_DIR=/usr/lib64/llvm8.0/lib/cmake/clang
+
+Alternatively,the location of the llvm-config script could be set.
+
+``cmake CMakeLists.txt   -DLLVMCONFIG=/usr/lib64/llvm7.0/bin/llvm-config``
+
+By default cmake will build some tests and use python for some simple tests.
+To disable tests use  ``-DBINDER_ENABLE_TEST=OFF``. To suggest python versions to perform tests with use options like
+``-BINDER_TEST_PYTHON_VERSIONS=2,3.5``.  Note that version 0 can be used to check the generated sources agains the reference sources.
+
```
lyskov, line 138: Could you please add separate chapter (probably `misc`?) at the end of documentation and outline steps needed to be taken to integrate with GH CI?

andriish:
Hi @lyskov , 
there is a bit updated version : https://github.com/RosettaCommons/binder/pull/107 , but I will answer here.

.github/workflows/centos7/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM centos:7.7.1908
+
+# Container image that runs your code
+#FROM alpine:3.10
```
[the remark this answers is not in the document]
andriish: Yes. Please note that the updated version was actually tested in a separate repository which contains only CI

https://github.com/andriish/binderCI

.github/workflows/centos7/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+yum -y install  clang clang-devel clang-libs  llvm-devel llvm-static
+yum -y install  libcxx-devel cmake3 make cmake gcc gcc-c++ 
+yum -y install yum-plugin-copr
```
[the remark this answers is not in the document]
andriish: I use only the official docker images from Fedora. It makes no sense to create and maintain own images.

.github/workflows/centos8/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -l
+set -x
+uname -a 
+cat /etc/issue
+yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+yum -y install  clang clang-devel clang-libs  llvm-devel llvm-static
+yum -y install  cmake make gcc gcc-c++ diffutils
```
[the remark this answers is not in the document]
andriish: gcc is the default compiler on RedHat. That is a good thing to test with the default compiller.
 Of course, the command 

 cmake -DCMAKE_CXX_COMPILER=clang++

 will work as well...

.github/workflows/centos8/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -l
+set -x
+uname -a 
+cat /etc/issue
+yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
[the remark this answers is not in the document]
andriish: Hm. The first line is about getting the extra packages if these are needed. Actually I'm not sure how the things are split in CentOS8. Maybe everything would work w/o EPEL, but one has to check that. In any case the EPEL will not hurt. There are no overlaps with the Base repositories.

CMakeLists.txt
```diff
@@ -1,20 +1,113 @@
 #This file can be used to build binder using preinstalled LLVM/Clang
-CMAKE_MINIMUM_REQUIRED(VERSION 3.4.3)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.7.0)
```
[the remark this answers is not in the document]
andriish: In this MR there are tests with python. The modules that search for python need quite recent cmake. However, these modules were excluded from the  updated merge request, so the merge reques has much lower requirements: 2.8+

.github/workflows/main.yml
```diff
@@ -0,0 +1,80 @@
+on: [push]
+
+jobs:
+  compilejobcentos7:
```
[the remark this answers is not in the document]
andriish: The updated merge request has better names.

documentation/install.rst
```diff
@@ -72,44 +72,46 @@ The basic dependencies for this type of installanion are very similar to these d
...
-
- ``yum install llvm-toolset-7.0* libcxx libcxx-devel``
-
- Then the compilation can be performed using the following shell
-
- ``scl enable llvm-toolset-7.0 bash``
-
-
-- For the Fedora 31+ the dependencies can be installed from the standard repositories. Run as root 
+On the RHEL7/RHEL8/Fedora22+/Ubuntu18+  systems binder can be compilled with the llvm, clang and dependent packages available
```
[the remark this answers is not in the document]
andriish: RHEL7 provides 3.4 by default. Yes, it works with 3.4.  The LLVMs from toolsets are of versions 5, 6 and 7. If I remember correctly.

source/CMakeLists.txt
```diff
@@ -33,15 +29,84 @@ add_clang_executable(binder
   fmt/format.cc
   fmt/posix.cc
   )
+if(NOT USE_EXTERNAL_LLVM)
```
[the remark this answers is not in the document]
andriish: Will do that.

source/CMakeLists.txt
```diff
@@ -33,15 +29,84 @@ add_clang_executable(binder
...
+target_link_libraries(binder
+  PRIVATE
+  clangTooling
+  clangBasic
+  clangASTMatchers
+  clangAST
+  clangFrontend
+  )
+else()
+find_library(LLVMSO  NAMES LLVM
```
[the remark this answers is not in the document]
andriish: That is just a name for variable that would hold a location of libLLVM.so

source/CMakeLists.txt
```diff
@@ -33,15 +29,84 @@ add_clang_executable(binder
...
+clangEdit 
+clangFormat 
+clangFrontend
+clangFrontendTool
+clangLex 
+clangParse
+clangRewriteFrontend
+clangSema
+clangSerialization
+clangTooling
```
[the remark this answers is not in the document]
andriish: Yes, the list is long. Everything was added one by one.

A better ordering is in the updated MR.

source/CMakeLists.txt
```diff
@@ -33,15 +29,84 @@ add_clang_executable(binder
...
+-Wl,--end-group
+)
+endif()
+endif()
```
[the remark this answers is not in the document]
andriish: The indentation is improved for the updated MR.

source/binder.cpp
andriish: OK.

source/binder.cpp
andriish: Ptr in the end of type implies that thas is  a pointer in a wide sense: raw, std::unique, std::shared, boost,  custom, etc.

Yes, it is in one place but it looks better than one mode ifdef.

source/binder.cpp
andriish: The absolutely lowest range I've tested is 3.4 just because it is the standard version in RHEL7 and I've started the porting from that version.  The versions between 4.0 and 7.0 don't need ifdefs. The versions above 7 and below 4.0 need some macros. The exact range should be up to you. I can only suggest to bind the versions of binder to the versions 
of "default" LLVM and define in docs what i "supported" and what "provided as is". But that is just a suggestion.

source/binder.cpp
andriish: ASTConsumerPtr is a raw pointer in some llvm 3.X.

source/binder.cpp
andriish: OK.

source/binder.cpp
andriish: errs() is not present in old llvm, I think.

test/diffbinder.cpp
andriish: Because 

- diff is Unix-only, i.e. not actually portable

- this utility has the same definition of exit codes everywhere

- one would have to play with sed/awk to ignore some patterns, while the implementation in C++ is trivial and portable.

documentation/install.rst
andriish: So far in the updated MR the Ci is split from the support of multiple LLVM versions. So that would make sense to add it in a separate MR.

.github/workflows/centos7/entrypoint.sh
andriish: EPEL does not contain a newer llvm, the default llvm 3.4 is used. EPEL is needed for a newer cmake, called  cmake3.

documentation/install.rst
```diff
@@ -72,44 +72,46 @@ The basic dependencies for this type of installanion are very similar to these d
...
+- If a newer or specific version of the llvm/clang is needed, it can be installed  as root
   
-  And run as root
+  ``yum install clang8.0 clang8.0-devel llvm8.0-devel llvm8.0-static clang8.0-libs``
+   
+  to obtain a specific version (8.0 in this case).
+    
+- If the option above is not sufficient, or the available packages are outdated, for the 
+  CentOS/RHEL/Fedora and compatible systems the llvm-toolset-7.0 toolset (or later) from
+  https://www.softwarecollections.org/en/scls/rhscl/llvm-toolset-7.0/ provides . Run as root
```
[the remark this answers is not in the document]
andriish: One can any of them. That is the point.  A typical Fedora user would not use software collection just because Fedora contains the newest versions of everything. On the THEL one could easily use the default but old  version, but if a more recent version of llvm would be needed, it should be taken from collections.

.github/workflows/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
andriish: I'm not an expert in github CI (prefer gitlab), but as far as I understand ubuntu is the docker host.  Github uses it as a primary linux platform for the obvious reasons, so I'm not even sure if RHEL@RHEL is available.

CMakeLists.txt
```diff
@@ -1,20 +1,113 @@
...
+set(LLVM_DIR_ORIG ${LLVM_DIR})
+set(Clang_DIR_ORIG ${Clang_DIR})
+set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules ${CMAKE_MODULE_PATH})
+SET(CMAKE_MODULE_PATH_ORIG "${CMAKE_MODULE_PATH}")
+set (CMAKE_MODULE_PATH "${LLVM_DIR};${Clang_DIR};${CMAKE_MODULE_PATH}")
+option(BINDER_ENABLE_TEST      "Enables building of tests" ON)
+option(BINDER_MOCK_TEST        "Fake running  binder on tests" OFF)
+option(BINDER_USE_PYTHON_IN_TEST  "Try to import python modules during tests" ON)
+if (NOT DEFINED BINDER_TEST_PYTHON_VERSIONS)
+set(BINDER_TEST_PYTHON_VERSIONS "0,2.7,3")
```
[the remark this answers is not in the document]
andriish: This is done in the updated MR.

source/binder.cpp
andriish: Resolved.

source/CMakeLists.txt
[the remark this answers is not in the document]
andriish: Resolved. See updated request #107

source/binder.cpp
andriish: Resolved. See updated request #107

documentation/install.rst
andriish: Would make sense to split the merge requests related to Ci as follows:

 - A  MR with CI that checks only if binder compiler. That is the setup from here
 https://github.com/andriish/binderCI. Optionally more OS or compilers can be added.
- A MR that implements a testing suite with ctest.  This is partially implemented in #96 
- A MR that brings the ctest testing to the CI.

test/diffbinder.cpp
lyskov: -- I see your point about `diff` being UNIX only, - good point!

-- re implementation in this case i think writing this in python would be much better approach and lead to considerably simpler code.

lyskov:
Ah, i was not aware about #107 i will close this and put future comments in #107.

.github/workflows/centos7/entrypoint.sh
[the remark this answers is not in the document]
lyskov: Hm... i think you missing my point: if we put install procedure into entry-point script it will be executed _each_time_container_run_. However if we put it into docker-file it will be run _only_once_ during container build. Obviously later will take much less resources.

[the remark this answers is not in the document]
lyskov: lets move all follow up on this into #98

.github/workflows/main.yml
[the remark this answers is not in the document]
lyskov: ah, ok - got it!

source/CMakeLists.txt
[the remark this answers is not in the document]
lyskov: in that case `lib_llvm_so_path` will be much better name.

source/binder.cpp
lyskov: You missing my point here: deleting unique pointer will lead to extra call to destructor which lead to double-free-error because API return raw pointer (ie no-owning-semantic) instead of unique-pointer (ie owner of allocated object)

test/diffbinder.cpp
andriish: Python makes sense as well. Despite it is not as portable as C++ and is much more painful to configure. But regexps are easirer.

.github/workflows/centos7/entrypoint.sh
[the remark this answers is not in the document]
andriish: OK. See #98

PR #98, opened 2020-03-15
Implementation of CI

andriish:
Dear @lyskov ,

here is an implementation of CI with the following features:
builds with external LLVM on push and a full builds with a recompilation  every 12 hours.
The first build uses Ubuntu18 as it is one of few systems that has llvm6 in the standard repositories. The second build uses CentOS7 (which is RH compatible, as you requested) but can be changed to something else.

Please note that all the changes affect only the build system, no changes to code were done.

In addition to that, the workflows directory contains some configurations for other OS which can be used in the tests.

Best regards,

Andrii


build.py
```diff
@@ -91,9 +91,17 @@ def install_llvm_tool(name, source_location, prefix, debug, jobs=1, clean=True,
...
+		execute('Download llvm source.', 'curl https://releases.llvm.org/{llvm_version}/llvm-{llvm_version}.src.tar.xz --output llvm-{llvm_version}.src.tar.xz'.format(llvm_version=llvm_version) )
+		execute('Decompress llvm source.', 'unxz  llvm-{llvm_version}.src.tar.xz'.format(llvm_version=llvm_version) )
+		execute('Decompress llvm source.', 'tar -xf  llvm-{llvm_version}.src.tar'.format(llvm_version=llvm_version) )
+		execute('Move llvm source.', 'mv llvm-{llvm_version}.src {prefix}'.format(llvm_version=llvm_version, prefix=prefix) )
+
+    if not os.path.isdir(clang_path): 
+		execute('Download clang source.', 'curl https://releases.llvm.org/{llvm_version}/cfe-{llvm_version}.src.tar.xz --output cfe-{llvm_version}.src.tar.xz'.format(llvm_version=llvm_version) )
+		execute('Decompress clang source.', 'unxz  cfe-{llvm_version}.src.tar.xz '.format(llvm_version=llvm_version) )
+		execute('Decompress clang source.', 'tar -xf  cfe-{llvm_version}.src.tar'.format(llvm_version=llvm_version) )
+		execute('Download clang source.', 'mv cfe-{llvm_version}.src {clang_path}'.format(llvm_version=llvm_version, clang_path=clang_path) )
```
lyskov, line 104: Could you please elaborate why this this change is needed? Particularly Original code used pipes to eliminate need for temporary files which will speed up things a bit.

cmake/Modules/CMakeMacroParseArguments.cmake
```diff
@@ -0,0 +1,138 @@
...
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of CMake, substitute the full
+#  License text for the above reference.)
```
lyskov, line 74: Please move "author" and license info to beginning of the file

test/diffbinder.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return EXIT_SUCCESS;
+}
+int main(int argc, char** argv)
+{
+    if (argc!=3) {
+        std::cout<<"Wrong number of arguments"<<std::endl;
+        return EXIT_FAILURE;
+    }
+    return COMPARE_ASCII_FILES(argv[1],argv[2]);
+}
```
lyskov, line 55: Why exactly this utility is needed?

cmake/Modules/CMakeFindFrameworks.cmake
```diff
@@ -0,0 +1,32 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
lyskov, line 1: here and other files in `Modules` dir: - this is _a lot_ of third-party-code and i would rather not include anything but absolutely essential code into project to reduce future maintenance cost. Could you please argue why it is needed?

Also, could you please rename to `modules` (no capitalization) so it in line with rest of directory structures

.github/workflows/centos7night/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM centos:7.7.1908
```
lyskov, line 1: Could you please argue why running nightly build for CentOS-7 make sense? It was released more then six years ago so it seems unlikely to me that suddenly some breaking changes will appear...

andriish:
Continuation of #96 
```
Hm... i think you missing my point: if we put install procedure into entry-point script it will be executed each_time_container_run. However if we put it into docker-file it will be run only_once during container build. Obviously later will take much less resources.
```

That would be true for a non-cloud system. But for the cloud one should compare things more carefully.

For the  solutions (let us call them A and B )

A:
Large custom docker image with preinstalled packages
B: 
Small standard docker image and installation of packages for each pull

The time to run A and B would not be much different. Why so?

It will take the same time to start docker

"B" will be faster to get the data, as it will be just a small docker image (maybe even cached) + yum (most likely from the local mirror or cached). "A" would require to pull the huge image from the Docker site (unlikely to be cached) + run yum update on top. It could be really slow.

"A" would not require any time to install packages as they are preinstalled. "B" would require maybe 1 minute to do it. 

The compilation should take the same amount of time.

So, in the end, the CI with the preinstalled packages has the potential to be ~1 minute faster, i.e. A would run in  5 minutes vs. B in  6 minutes. Maybe. And only in the case when there are no updates, i.e. for unsupported systems.
(Right now the execution time of B configuration is between 2 minutes and 6 minutes 19 seconds. See this repo https://github.com/andriish/binderCI )

So far so good. But the maintenance of the configuration  A would cost a lot of manpower if to compare the potential improvement of 1 minute run time for some cases.  Even if one should spend unrealistically 10-15 minutes per Docker image per year  ( one should keep the system updated!)  that adds up to some hours even for couple images. 
It also looks like a bad deal to trade yours or mine time for already paid CPU minutes of the GitHub/Microsoft. 

So well, yes, I think the whole point of the docker is to avoid building own environments.

.github/workflows/centos7/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM centos:7.7.1908
+
+# Container image that runs your code
+#FROM alpine:3.10
+
+# Copies your code file from your action repository to the filesystem path `/` of the container
+COPY entrypoint.sh /entrypoint.sh
+
+# Code file to execute when the docker container starts up (`entrypoint.sh`)
+ENTRYPOINT ["/entrypoint.sh"]
```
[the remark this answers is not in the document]
andriish: Please see the answer below.

build.py
andriish: OK.  For some reason, the xz format is not supported on some systems in the tar and there is a separate unxz utility.

.github/workflows/centos7/Dockerfile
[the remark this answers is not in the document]
andriish: Also, this PR is kind of obsolete as well -- the updated version of Ci is located here  
https://github.com/andriish/binderCI
and once the issues are discussed I will create a new MR.

cmake/Modules/CMakeMacroParseArguments.cmake
andriish: I would prefer to avoid modifications. Not because of any copyright (it is BSD)  but because this file is almost standard for cmake.  In this respect, it is similar to configure.guess from autotools or similar files.   People just use standard versions "as is".

test/diffbinder.cpp
andriish: To compare the output of the compiled executable to the reference.

cmake/Modules/CMakeFindFrameworks.cmake
andriish: Content:
These files are just equivalents of m4 files for autotools + some macros.
Most are from the cmake.
The typical structure of cmake project contains the 'cmake' directory with 
cmake/Modules   <- here one has files that search for dependencies. Typically needed to search for SW that is not very common and has no cmake configs.
cmake/Templates  <-  configs to be installed for this project

Name: 
That comes as a standard from cmake as well.  
For instance I have '/usr/share/cmake/Modules' with capital M.

.github/workflows/centos7night/Dockerfile
andriish: The idea is to test binder and not CentOS7. Concerning the later: it is the production system and will be supported for the next 5-10 years. In our field 90%+ cluster machines run it.   Of course, you are right,  there are no incompatible changes and most of the things get only the most crucial updates, but for instance, cmake is updated on a regular basis. I have 3.17.1 now.

andriish:
> @andriish thank you for splitting this into separate PR! This makes things much more manageable. In addition to comments below could you please add documentation how to enable GH CI? Thanks,

Yes. Will document it.

andriish:
Send an invitation to  binderCI just so you will see the actions and the timing.

build.py
andriish: Please note this change is not present in the binderCI repo now.

andriish:
Right now the CI in https://github.com/andriish/binderCI  is triggered on commits and checks compilation success. It can be put in this form  into the upstream  (the lines like git check .. will be removed, of course)  once the desired range of the LLVM versions is defined. Once it is clear, I will prepare a MR.
The regular builds as well as running of tests via cmake can be done afterwards.

.github/workflows/centos7night/Dockerfile
lyskov: Oh, i am not arguing of usability of CentOS-7, - it the same here, - all major clusters is using it now and will use it for quite a while. My point was that it is very stable environment so nightly tests seems a bit excessive. But i see your point of this being Binder "nightly" tests, (not the CentOS's one)

lyskov:
Bumping this since #108 was merged. Thanks,

andriish:
I've created https://github.com/RosettaCommons/binder/pull/109  with some description of features and possible extensions.
Concerning the documentation:  there are no special things inside (so far) and if someone understands how the CI works in general the workflow of this particular CI looks trivial.  So "code is the best documentation" does not look like a joke for this case. Some less trivial things are commented online...

andriish:
OK. Close this one. Some parts will be done in other merge requests.

PR #104, opened 2020-05-02
Implement option for output in multiple files with fixed name.

andriish:
Dear Sergey,

if you will have some time, here is a pull request for your consideration.

1) The pull request implements an option --multiple-files=n that makes binder to write the output into  files with names:  root-module-name + '.cpp' and  root-module-name + '_k.cpp', k=1..n. Basically that is the same as output with --single-file option but the sources  split into n pieces.

This is useful because:

 For some codebases of intermediate size having all bindings in a single file is not an option. 
 binder --single-file produces huge files. The compilation cannot be parallelized, it takes too much real-time to compile them and too much memory.

Using binder in a normal mode works fine, but the names of the produced files, their number and size vary depending on the used headers and binder setup.  This is not very practical when these files are stored in git or svn.

Binder with the implemented option produces multiple files with an approximately equal number of lines and the same names regardless of the used setup, solving both problems described above.
It works best with  --max-file-size set to some quite low number.

2) The codes were tested with the "standard" way of build with LLVM60
For the codes I was testing the option has allowed a significant reduction of real compilation time of produced bindings (~2min40s -> 40s) and a reduction of the peak memory consumption by gcc   (5.3Gb -> 1Gb) in comparison to the single file option.  

Best regards,
Andrii

### DIFF
--- a/source/binder.cpp
+++ b/source/binder.cpp
@@ -70,7 +70,7 @@ cl::opt<bool> O_annotate_includes("annotate-includes", cl::desc("Annotate each i
 
 cl::opt<bool> O_single_file("single-file", cl::desc("Concatenate all binder output into single file with name: root-module-name + '.cpp'. Use this for a small projects and for testing."), cl::init(false), cl::cat(BinderToolCategory));
 
-cl::opt<int> O_single_file("multiple-files", cl::desc("Similar to the single-file option, but the output is split into n pieces and written into files with names: root-module-name + 'k.cpp', k=1..n."), cl::init(0), cl::cat(BinderToolCategory));
+cl::opt<int> O_multiple_files("multiple-files", cl::desc("Similar to the single-file option, but the output is split into n pieces if possible and is  written to files with names: root-module-name + '_k.cpp', k=1..n."), cl::init(0), cl::cat(BinderToolCategory));
 
 cl::opt<bool> O_trace("trace", cl::desc("Add tracer output for each binded object (i.e. for debugging)"), cl::init(false), cl::cat(BinderToolCategory));
 
--- a/source/binder.cpp
+++ b/source/binder.cpp
@@ -70,6 +70,8 @@ cl::opt<bool> O_annotate_includes("annotate-includes", cl::desc("Annotate each i
 
 cl::opt<bool> O_single_file("single-file", cl::desc("Concatenate all binder output into single file with name: root-module-name + '.cpp'. Use this for a small projects and for testing."), cl::init(false), cl::cat(BinderToolCategory));
 
+cl::opt<int> O_single_file("multiple-files", cl::desc("Similar to the single-file option, but the output is split into n pieces and written into files with names: root-module-name + 'k.cpp', k=1..n."), cl::init(0), cl::cat(BinderToolCategory));
+
 cl::opt<bool> O_trace("trace", cl::desc("Add tracer output for each binded object (i.e. for debugging)"), cl::init(false), cl::cat(BinderToolCategory));
 
 cl::opt<bool> O_verbose("v", cl::desc("Increase verbosity of output"), cl::init(false), cl::cat(BinderToolCategory));
--- a/source/binder.hpp
+++ b/source/binder.hpp
@@ -19,6 +19,7 @@
 
 extern llvm::cl::opt<bool> O_annotate_includes;
 extern llvm::cl::opt<bool> O_single_file;
+extern llvm::cl::opt<int>  O_multiple_files;
 extern llvm::cl::opt<bool> O_trace;
 extern llvm::cl::opt<bool> O_verbose;
 
--- a/source/context.cpp
+++ b/source/context.cpp
@@ -458,9 +458,9 @@ void Context::generate(Config const &config)
 
 			code = generate_include_directives(includes) + fmt::format(module_header, config.includes_code()) + prefix_code + "void " + function_name + module_function_suffix + "\n{\n" + code + "}\n";
 
-                        if( O_multiple_files>0 ) {
-                        outputs.push_back(std::pair<string,string>(file_name,code));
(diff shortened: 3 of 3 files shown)


lyskov:
Thank you for putting this together @andriish ! Having stable output is indeed an issue. Could you please elaborate how this solution is better then using current file-naming-schema when `--max-file-size` is set to `0`? (this is what i use for mine own projects when i need stable output). Thanks,

lyskov:
Also, please note that initial reason why i choose "tree like" structure for storing generated files instead of "flat" schema proposed here is because in "tree" version any "overflow changes" (when generated code suddenly does not fit into same file) is very isolated. When in the "flat" schema such changes when then happened at lower indexes will lead to entirely different files set (ie disproportionally large number of files will got changed)

andriish:
Dear @lyskov ,

Starting from the second question: I totally agree that the tree schema is the only reasonable schema for a large project. The implemented option would be useful for small to medium size projects, so in my answer to your first question I will not assume a huge project.

So, coming back to the first question:
The main advantage is for the build system and VCS.

With the --max-file-size=0**  not only the content of the generated files, but their names and number of these files will depend on the a)used binder settings and b) the content of the binded sources/headers.  That has two implications:  
a) that is more tricky for the build system to know what to build. Of course something like cmake can coupe with it easily, binder even writes a list of source files, etc., but for other build systems it can be tricky.
b) If one would like to keep the generated binder sources in VCS, it is necessarily to keep track of the generated  sources. Means that after each change to the binder config one would have to add/remove manually the files from VCS. In some cases these changes would be needed even after an update of the binded headers, which is not an obvious thing.

So the option of multiple files solves the problems of a) and b).  The names and the number of the produced source files will be always the same, regardless of the binder config and the binded classes.

Of course, the option does not guarantee the *content* of the generated files will be similar and in this respect the tree structured output wins.  But for a small to medium size project with 10-50 files produced by binder it is completely OK even if all the files will be changed. It is more important to have a well-defined output names for the files.


Best regards,

Andrii

** Actually the --max-file-size=0  from  upstream/master does not work for me for some reason. 
It just hangs and produces thousands of files.  But I just assume what it should do  is the same as 
--max-file-size=some_huge_number, right? If that is so, then multiple files option has an extra advantage: it can split files into small chunks keeping the names simultaneously consistent.

lyskov:
Thank you for detailed explanation @andriish ! I see your point of having "flat" schema for generated files, - i guess this could be useful for small projects. (I am not sure however if i understand what you mean by need to manually add generated sources to VCS though, - it is trivial to script addition/deletion of generated files to keep VCS up-to-date).

Also, using total file number as limit have the same problem as using single file, - sooner or later generated files will became too large and manual intervention will be needed. So i think using max-file-size is better metric here. Also, specifying number of files is problematic for cases when total number generated code blocks is less then requested number of files (we literally will need to generate empty files in that case).

I just pushed e39c2dd which add preliminary implementation of `--flat` option which tell Binder to save files into single directory instead of writing output files in tree-like structure. This implementation does not try to squash multiple code blocs into single file though. Also, current implementation (and one in this PR) is suboptimal in a sense that it tries to pack code blocks _after_ include directives was generated. This will create unnecessary large and hard-to-read files with multiple include blocks. Fixing this however will require a bit larger refactoring. I will see to it (and to how even file sizes) later.

andriish:
Hi @lyskov,

thanks a lot! That option makes things much simpler!

1) With adding/removing from VCS I mean that one was not able to predict number and names of files and after updates of  binder config or other files, a cleanup would be needed, e.g.

version N of VCS has

A_1.cpp A_2.cpp B.cpp  D_1.cpp D_2.cpp generated by binder in tree more

but  after the update of binder config the generated files in tree mode  would be:

A.cpp B_1.cpp D_1.cpp D_2.cpp E.cpp

So the files from version N should be removed and the generated files A.cpp B_1.cpp D_1.cpp D_2.cpp E.cpp   should be added to get a proper version N+1.

Of course, that is scriptable, but quite bug prone.

2)  The flat scheme. With the tree scheme the names of output files are completely unknown before the binder execution, which makes writing makefile rules hard.
Now this problem is a bit simpler. The files generated with --flat always have  well-defined names, e.g. 
A_1.cpp A_2.cpp  ...     A_X.cpp.
And after a significant update the only change would be the number of files, e.g.
A_1.cpp A_2.cpp  ...     A_Y.cpp.
The number of files Y  is still  not known before the execution of binder, but at least some ugly workarounds are possible.  And here is the thing concerning the empty files. 
You are right, in the scheme with fixed number of files, in some cases there will be empty files. But that is kind of OK, an empty file rise no compilation errors. And  exactly that would allow some  workarounds to make the flat scheme look almost indistinguishable from the scheme with a fixed number of files.

In terms of GNU  Makefile the flat scheme usage would be:

mymodule.so: A_1.cpp A_2.cpp  ....   A_100.cpp
                $(CXX) ....
A_1.cpp A_2.cpp  ....   A_100.cpp:   mymodule.config
         rm -f  A_1.cpp A_2.cpp  ....   A_100.cpp
          touch A_1.cpp A_2.cpp  ....   A_100.cpp  # create a lot of empty files, for sure more than needed 
          binder -flat=1 --root-module=A .....  mymodule.config  #overwrite some empty files

The fixed number of files scheme would be

mymodule.so: A_1.cpp A_2.cpp  ....   A_20.cpp
                $(CXX) ....
A_1.cpp A_2.cpp  ....   A_20.cpp:   mymodule.config
          binder -multiple-files=20 --root-module=A .....  mymodule.config  # we always know the output.  Some files could be empty, but we don't care.

So in the end the fixed number of files scheme has this one advantage -- completely predictable output.


3) Suboptimal output -- yes. But that is hard to fix, true.

I hope I was able to explain the things clear.

Andrii

PR #105, opened 2020-05-04
WIP: Remove special characters from function names

### LINKED ISSUE #103: Hyphen-minus ("-") in file names produces invalid C++ identifiers
Hello!

I'm trying to generate bindings for a C++ library where some folder and file names contain the "-" character. Based on the file names Binder produces invalid C++ identifiers.

Example:
When I try to bind code from a `demo-project` folder and my header file is called `demo-1.h`, generation of bindings succeeds without errors, but building attempt results in:

```
sourcedemoproject.cpp:11:15: error: expected initializer before ‘-’ token
 void bind_demo-project_demo-1(std::function< pybind11::module &(std::string const &namespace_) > &M);
               ^
sourcedemoproject.cpp: In function ‘void pybind11_init_demoproject(pybind11::module&)’:
sourcedemoproject.cpp:33:2: error: ‘bind_demo’ was not declared in this scope
  bind_demo-project_demo-1(M);
  ^~~~~~~~~
sourcedemoproject.cpp:33:12: error: ‘project_demo’ was not declared in this scope
  bind_demo-project_demo-1(M);
            ^~~~~~~~~~~~
sourcedemoproject.cpp:33:28: error: expression cannot be used as a function
  bind_demo-project_demo-1(M);
```

I'm guessing this is a known issue, but I couldn't find it here.
Is there a workaround other than renaming/editing files?

andriish:
Dear @lyskov ,

this is a simple attempt to fix #103. The code has not been well tested yet, 
but hopefully it can be useful.  

Andrii

### DIFF
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
 	return r;
 }
 
+/// Move all occurances of non-alphanumerical characters to the end of the string and replace them with their ASCII codes.
+std::string to_alphanumerical(const std::string& input)
+{
+    std::string output;
+    std::string suffix;
+    size_t pos=0;
+    for (auto c: input) {
+    
+        if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
+            output+=c;
+        else
+            { suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(int(c)); }
+     pos++;
+     }
+    return suffix.size()?output+"fixed"+suffix:output;
+}
+
 
 /// check if string begins with given prefix
 bool begins_with(std::string const &source, std::string const &prefix)
--- a/source/util.cpp
+++ b/source/util.cpp
@@ -74,20 +74,19 @@ string replace_(string const &s, string const & from, string const &to)
 }
 
 /// Move all occurances of non-alphanumerical characters to the end of the string and replace them with their ASCII codes.
-std::string to_alphanumerical(const std::string& input)
+std::string to_alphanumerical(std::string const & input)
 {
-    std::string output;
-    std::string suffix;
-    size_t pos=0;
-    for (auto c: input) {
-    
-        if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
-            output+=c;
-        else
-            { suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(int(c)); }
-     pos++;
-     }
-    return suffix.size()?output+"fixed"+suffix:output;
+	std::string output;
+	std::string suffix;
+	size_t pos=0;
+	for (auto c: input) {
+		if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
+			output+=c;
+		else
+		{ suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(c); }
+		++pos;
+		}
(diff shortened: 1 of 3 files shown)


source/util.cpp
```diff
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
...
+    
+        if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
+            output+=c;
+        else
+            { suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(int(c)); }
+     pos++;
+     }
+    return suffix.size()?output+"fixed"+suffix:output;
+}
+
```
lyskov, line 92: why i am not directly oppose moving suffixes to the end but this strategy may seem a bit confusing to user, - why not _replace_ characters with ascii codes instead? Moreover, i am guessing that you want to add ASCII codes to make function names unique? If so then this is not fully "bullet proof" strategy since it is possible that two includes exist: one with special characters and another with numeric suffixes. To address this better would be to convert to alphanumeric at name-prefix level (ie line ~376 of context.cpp) and allow current code to handle duplicates for us. This will also lead to much more readable file names.

source/util.cpp
```diff
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
...
+    std::string output;
+    std::string suffix;
+    size_t pos=0;
+    for (auto c: input) {
+    
+        if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
+            output+=c;
+        else
+            { suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(int(c)); }
+     pos++;
```
lyskov, line 88: please do not use use post-increment, while compiler may optimize this technically this is sub-optimal. Please see https://isocpp.org/wiki/faq/operator-overloading#increment-pre-post-speed

source/util.cpp
```diff
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
 	return r;
 }
 
+/// Move all occurances of non-alphanumerical characters to the end of the string and replace them with their ASCII codes.
+std::string to_alphanumerical(const std::string& input)
```
lyskov, line 77: please use the same style of `const` placement as rest of the code

source/util.cpp
```diff
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
...
+{
+    std::string output;
+    std::string suffix;
+    size_t pos=0;
+    for (auto c: input) {
+    
+        if ( ( c>='0' && c<='9') ||  ( c>='a' && c<='z') || ( c>='A' && c<='Z') ||  c=='_' )
```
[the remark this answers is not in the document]
andriish: That would be isalnum + '_'. isalnum without locale is de-facto a C function anyway.
So this is a more explicit version.

source/util.cpp
```diff
@@ -73,6 +73,23 @@ string replace_(string const &s, string const & from, string const &to)
...
+            output+=c;
+        else
+            { suffix+='_'; suffix+=std::to_string(pos); suffix+='_'; suffix+=std::to_string(int(c)); }
```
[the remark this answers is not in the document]
andriish: int(c) stands here more for readability. OK.

source/util.cpp
andriish: That is not important for integer types, but yes.

source/util.cpp
andriish: Oh, yes.

andriish:
And the formatting -- the code was formatted  automatically with 'astyle', i.e. no tabs. I've put tabs manually.

lyskov:
actually never mind, -  i will push fix for this issue with another commit later today. Thanks,

lyskov:
FYI, fixed in 170978f

PR #108, opened 2020-06-10
Multiple llvm versions 6+

andriish:
Support of standalone LLVM 6+.
De-facto that is 4+, as there is no difference between versions 4/5/6.


source/binder.cpp
```diff
@@ -13,6 +13,8 @@
 #include <binder.hpp>
 
 // Declares clang::SyntaxOnlyAction.
+#include "clang/AST/ASTConsumer.h"
+#include "clang/AST/ASTContext.h"
```
lyskov, line 17: Could you please double check if these includes should not be need here? Or is it due to high version of LLVM  changes?

source/binder.hpp
```diff
@@ -15,6 +15,7 @@
 #define _INCLUDED_binder_hpp_
 
 
+#include "llvm/Config/llvm-config.h"
```
lyskov, line 18: Optional: again, is this include really needed or is this result of removed LLVM support?

source/CMakeLists.txt
```diff
@@ -33,6 +33,48 @@ add_clang_executable(binder
...
+                           LLVM-${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}
+                           PATHS  ${LLVM_LIBRARY_DIR}  ${LLVMLIBDIR}
+                           )
+add_definitions(${LLVM_DEFINITIONS})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${LLVM_COMPILE_FLAGS}")
+list(REMOVE_ITEM CMAKE_CXX_FLAGS "-fno-exceptions")
+string (REPLACE "-fno-exceptions" ""     CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+message(STATUS "binder: LLVMSO=${LLVMSO}")
+message(STATUS "binder: CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}")
+if ( (LLVM_VERSION_MAJOR GREATER 3)  AND (LLVM_VERSION_MAJOR LESS 8) )
```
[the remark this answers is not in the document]
andriish: There should be something in the with "greater" . But as 4 and 5 are technically the same as 6, 3 is therem

source/binder.cpp
lyskov: 👍

source/CMakeLists.txt
```diff
@@ -33,6 +33,48 @@ add_clang_executable(binder
...
+  )
+endif()
+if ( (LLVM_VERSION_MAJOR EQUAL 10) OR (LLVM_VERSION_MAJOR GREATER 10) )
+target_link_libraries(binder
+  PRIVATE
+  ${lib_llvm_so_path}
+  clang
+  clang-cpp
+  )
+endif()
```
lyskov, line 76: please fix indentation in this block

source/CMakeLists.txt
```diff
@@ -42,6 +84,7 @@ target_link_libraries(binder
   clangAST
   clangFrontend
   )
+endif()
```
lyskov, line 87: Please fix indentation in line range 78-86. Thanks,

source/binder.hpp
lyskov: Ah, i see it now, - the lack indentation confused me. Please see two comments regarding it.

source/CMakeLists.txt
andriish: Done for the whole file.

lyskov: 👍

PR #109, opened 2020-06-27
Linux CI: fast version with external LLVM using Fedora 

andriish:
Hi @lyskov  ,

this MR implements CI for Linux.

Note:
 - while the CI now attempts to run tests via cmake, the result of CI job is defined only with the result of the compilation.
That is why there is a third step in the CI.  Please note that the existing tests are not portable across compilers and platforms, so some logic should be added to the test suite before actually use the results of tests to define the outcome of the CI jobs. 

 - CI runs on push/pull and every 5 days on cron. Af far as I remember this works only on master branch.
 - Windows and mac can be added as well, but it makes sense to do that in another MR.
 - A version with full rebuild can be added as well, but again it makes more sense to have a separate MR for that.
 - There are many docker images from other distributions, so that would make sense to use some that are widespread in your community.  I can add some generic of redhat7/8 later, as that is used in HEP.

Best regards,
Andrii

### DIFF
--- a/.github/workflows/fedora26/Dockerfile
+++ b/.github/workflows/fedora26/Dockerfile
@@ -1,0 +1,8 @@
+# Container image that runs your code
+FROM fedora:26
+
+#Copies your code file from your action repository to the filesystem path `/` of the container
+COPY entrypoint.sh /entrypoint.sh
+
+#Code file to execute when the docker container starts up (`entrypoint.sh`)
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/workflows/fedora26/action.yaml
+++ b/.github/workflows/fedora26/action.yaml
@@ -1,0 +1,15 @@
+name: 'Compile binder'
+description: 'Binder'
+inputs:
+  theextraargument:   # an optional extra argument for docker run.
+    description: 'Extra docker argument'
+    required: true
+    default: 'World'
+outputs:
+  thestatus: # here we put the execution status 
+    description: 'The execution status'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
+  args:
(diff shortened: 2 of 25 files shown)


andriish:
Hi @lyskov  ,

- Not sure what you mean in your first point.
  In case that is about the CI activation, it activates itself automatically if  (for the Actions CI) there is a config in the .github/workflows  directory.
  In case that is about the tests (compilation, comparison, execution) of binder test programs --  please see the next point about the implementation in binder. To use the  test results in CI one should do

```
cmake CMakeLists.txt
make
ldd source/binder
ldd -u -r source/binder
ctest . --output-on-failure 
out=$?
echo ::set-output name=out::$out
```
instead of 
```
cmake CMakeLists.txt
make
out=$?
ldd source/binder
ldd -u -r source/binder
ctest . --output-on-failure 
echo ::set-output name=out::$out
```

- Sure that is possible. It is partially implemented in the  #98. That is quite good (but not completed) implementation that relies only on cmake.  It works as follows:

    - cmake looks for pybind11 headers and (multiple)  Python libraries/headers.
    - cmake runs the binder and creates bindings for the tests
    - cmake compiles a portable  custom diff utility that can ignore some patterns in the inputs
    -  for each test X cmake creates two tests. The first one  checks the generated source X versus the reference. The second one runs the compiled module. More tests for the X can run other python versions.  
    - if any of the tests fail, the whole ctest is declared as failed.       
    - comparison vs reference sources or the actul execution of python modules can be disabled.

However, the implementation in #98 was done only for some test in the suite. The complete implementation requires a bit of work.

- The #98  a bit older version of this MR   but includes an implementation of the cmake test.  As for me that  makes sense to have these things in another MR.

PR #113, opened 2020-08-01
doc/install: Fix missing backticks

EricCousineau-TRI:


EricCousineau-TRI:
Currently looks like this:

> image

PR #116, opened 2020-08-01
doc/config: Use code formatting for flags

EricCousineau-TRI:
Rather than italics


EricCousineau-TRI:
Pics or it didn't happen:
> image

PR #119, opened 2020-08-03
test: Update reference generated files with --accept

### LINKED ISSUE #117: Tests fail on Ubuntu 18.04?
When I ran `./build-and-run-tests.py` on Ubuntu 18.04 (using master @ 022cd0d), I get the following errors:
https://gist.github.com/EricCousineau-TRI/4bd4615baeabc2c3b071473f597cbae5

FTR, the regression test errors seem pretty small, like something about an `enum class`?

EricCousineau-TRI:
Resolves #117

Generated (blindly) with the following setup on Ubuntu 18.04:

```sh
$ git log -n 1 --oneline --no-decorate
022cd0d Update README.md
$ mkdir build && cd build
$ cmake .. -DLLVM_DIR=/usr/lib/llvm-10 -DClang_DIR=/usr/lib/llvm-10
$ make -j 4
$ git clone https://github.com/pybind/pybind11
$ ( cd pybind11 && git log -n 1 --oneline --no-decorate; )
1caf1d0 fix: check list validity and add missing file (#2352)
$ cd ..
$ yes | ./test/self-test.py --accept --binder ${PWD}/build/source/binder --pybind11 ${PWD}/build/pybind11/include
```

### DIFF
--- a/test/T42.stl.names.ref
+++ b/test/T42.stl.names.ref
@@ -2,19 +2,11 @@
 #include <functional> // std::binary_function
 #include <functional> // std::equal_to
 #include <functional> // std::less
-#include <iterator> // __gnu_cxx::__normal_iterator
-#include <map> // std::_Rb_tree_color
-#include <map> // std::_Rb_tree_const_iterator
+#include <list> // std::_List_node
 #include <memory> // std::allocator
 #include <memory> // std::allocator_traits
-#include <memory> // std::pointer_traits
-#include <set> // std::multiset
-#include <set> // std::set
 #include <sstream> // __str__
-#include <string> // std::basic_string
-#include <string> // std::char_traits
 #include <utility> // std::pair
-#include <utility> // std::tuple_element
 
 #include <pybind11/pybind11.h>
 #include <functional>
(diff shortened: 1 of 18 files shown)


test/T42.stl.names.ref
```diff
@@ -2,19 +2,11 @@
 #include <functional> // std::binary_function
 #include <functional> // std::equal_to
 #include <functional> // std::less
-#include <iterator> // __gnu_cxx::__normal_iterator
-#include <map> // std::_Rb_tree_color
-#include <map> // std::_Rb_tree_const_iterator
+#include <list> // std::_List_node
```
lyskov, line 5: Most changes to this file is OS/libc++ dependent so results here is different from what i see on CentOS (our reference platform). Could you please revert all changes to this file except the removal of `pybind11::handle cl_type = cl;` and following new line? (if that too much trouble then revert the whole file and i will commit reference result from CentOS run later)

EricCousineau-TRI:
Lemme briefly try out the Dockerfiles in `.github` and see if it's uber easy to use there. If it takes more than 2min, I'll letcha know!

EricCousineau-TRI:
Took more than 2min :(

Used the `fedora33` as a starting point, tried running the test, but go the following error about missing libc headers, I think:
```
# yes | ./test/self-test.py --accept --binder ${PWD}/build/source/binder --pybind11 ${PWD}/build/pybind11/include
...
/usr/lib/gcc/x86_64-redhat-linux/10/../../../../include/c++/10/bits/cxxabi_init_exception.h:38:10: fatal error: 'stddef.h' file not found
/usr/lib/gcc/x86_64-redhat-linux/10/../../../../include/c++/10/bits/cxxabi_init_exception.h:38:10: fatal error: 'stddef.h' file not found
#include <stddef.h>
         ^~~~~~~~~~
1 error generated.
Error while processing /binder/test/build//T02.function.hpp.include
```

Can I ask if you can generate it on your machine?

andriish:
Hi @EricCousineau-TRI ,

that is a known feature of the clang tools.

One should add  -iwithsysroot<whereyouhave theclangincludes> after -- option.
As far as I understand, this happens when there is gcc installation in the system and the c++ headers from g++ take precedence over these from clang.

Please see this file for the full example: 
https://github.com/andriish/binder/blob/use_ctest/test/CMakeLists.txt

Lines 17-20


Best regards,

Andrii

lyskov:
@EricCousineau-TRI no worries, let me just regenerate this on my system so you do not need to struggle with an extra options...

lyskov:
@EricCousineau-TRI i pushed the update (see https://github.com/RosettaCommons/binder/commit/c8a86783187093698d516b5804b2e053655325cd) so this change set now is in master. - thank you for your help!

EricCousineau-TRI:
Awesome, thanks @lyskov and @andriish!

EricCousineau-TRI:
@andriish Just to check, should the `-iwithsysroot` be part of either `build-and-run-tests.py` or `self-test.py`?

andriish:
Hi @EricCousineau-TRI ,


it should be  in ``self-test.py``, around the line 

```
  command_line = '{binder}  ...
```
In general, I'm not very familiar with the way clang searches the paths, so I suspect that in some cases, ``-iwithsysroot`` will not be needed  and binder will work w/o it, but "explicit is better than implicit".  

Best regards,

Andrii

Issue #120, opened 2020-08-03
Tests don't seem like they're run on CI?


EricCousineau-TRI:
Split off from #117

When I look at CI, it doesn't look like the tests are run? (they assume CTest, but there are no CTest targets configured?)
For example:
https://github.com/RosettaCommons/binder/runs/896230166#step:3:1076
```
+ ctest . --output-on-failure
Test project /github/workspace
No tests were found!!!
```

andriish:
Hi @eacousineau ,

The ctest variation is under development.
 Therefore, the ctest command  is present in CI but the CI does not report a failure if the ctest command fails or no ctests found.
There will be a merge request with an implementation of ctest soon.

Best regards,

Andrii

andriish:
This can be closed, I think.

EricCousineau-TRI:
SGTM!

PR #121, opened 2020-08-06
Implementation of cmake tests (ctest)

andriish:
Hi @lyskov ,

 here is the MR we descussed in the past.

In this MR:
  - Implementation of cmake tests (ctest)
  - Usage the ctests in CI

How to use the ctest:

```
 cmake -BINDER_ENABLE_TEST=ON ....
 make  # or whatever is used 
 ctest .
```
The testing is fully enabled by default if:    cmake version is 3.0.0 or higher, pybind11 and at  python libraries/headers/interpreters are available.  Otherwise some tests will not be executed, e.g. obviously python3 tests will not be enabled if the Python3 libraries are not found.

Features:
  - The  update does not interfere with the existing python tests.

  - Multiple python versions can be and should be used in parallel. The versions are set as a comma-separated list passed by the ``BINDER_TEST_PYTHON_VERSIONS`` option. The default list is 0,2,3.
    The Python version "0" corresponds to a "plain diff" of the output vs reference. The versions of python2/Python3 will be the first     versions found by cmake.   For a specific versions one can use e.g. 
   ``   cmake  ....   -DBINDER_TEST_PYTHON_VERSIONS=0,2.7.15,3.8.0,3.7.0`` .  The generated codes will be compiled and loaded using the corresponding interpreter.

- With an option  ``-DBINDER_MOCK_TEST`` one can mock the code generation by binder. In this case the reference codes will be used in the python tests.

 - The actual comparison of the source codes is done not via GNU/Linux diff, but via simple "diffbinder" code in the repo. This assures portability (e.g. on Windows and Unixes with non-GNU diff like Solaris ) and allows to ignore certain patterns in the generated codes w/o an invocation of external programs that should treat regexps.

-  So far different versions of LLVM produce slightly different codes, so all the comparisons are run with an argument --always-success, which forces the diffbinder utility to report "success" regardless of the actual differences.  The actual difference is, of course still visible, e.g. in logs or using  ``ctest . --verbose ``

Technical things to note: 
  -  The  ``/cmake``  directory is added with some  cmake scripts from the standard cmake distribution. Some huge files, but these are standard.

Possible future extensions:
 - Include PyPy support

General observations for binder from this MR:
 - In case binder is compiled out of the LLVM source tree and the host system has installed gcc as a default compiller, one has to 
set the path to the CLang headers as an extra argument in the command line.   It would be nice to describe it in the documentation.

Tagging  @EricCousineau-TRI and link #120 

Best regards,

Andrii


CMakeLists.txt
```diff
@@ -14,6 +14,14 @@ set(Clang_DIR_ORIG ${Clang_DIR})
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules ${CMAKE_MODULE_PATH})
 SET(CMAKE_MODULE_PATH_ORIG "${CMAKE_MODULE_PATH}")
 set (CMAKE_MODULE_PATH "${LLVM_DIR};${Clang_DIR};${CMAKE_MODULE_PATH}")
+option(BINDER_ENABLE_TEST      "Enables building of tests" ON)
+option(BINDER_MOCK_TEST        "Fake running  binder on tests" OFF)
```
lyskov, line 18: Description should probably be ~"Mock running  binder..."

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
+#include <iostream>
+#include <fstream>
+#include <stdio.h>
+#include <vector>
+#include <string.h>
+
+int skip(const std::string& s1)
```
lyskov, line 7: Here and below: could you please use the same `const` notation as rest of Binder code? (if you curious, please note that there is a strong technical reason for preferring this notation, i think this comment explains well https://stackoverflow.com/a/5503602/7243597)

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
+#include <iostream>
```
lyskov, line 1: Could you please add author and copyright header

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+    if (s1.find("Source list file")!=std::string::npos) return 3;       //Same as the list of source files
+    if (s1.find("Modules list file")!=std::string::npos) return 4;      //Same as the list of modules
+    if (s1.find("File:")!=std::string::npos) return 5;                  //Same as the full file paths.
+    if (s1.find("pybind11::handle cl_type = cl;")!=std::string::npos) return 6;  //This could be optimized in binder
+    if (s1.find("__cxx11")!=std::string::npos) return 7;                         //This looks like implementation dependent
+    if (s1.find("assign")!=std::string::npos) return 8;                          // The assing is binded differently for some reasons
+    return 0;
+}
+int COMPARE_ASCII_FILES(const std::string& f1,const std::string& f2)
```
lyskov, line 23: Why capital letter for this function name? please justify? Also, name itself sounds rather confusing to me. Have you meant something like `compare_files_as_text` or `compare_text_files`?

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    std::fstream file1(f1.c_str()), file2(f2.c_str());
```
lyskov, line 25: Since we expect at least C++11 to be present then this should be `std::fstream file1(f1), file2(f2);`

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
...
+    std::string string1, string2;
+    int j1,j2;
+    j1 = 0;
+    j2 = 0;
+    if (!file1.is_open()|| !file2.is_open())
```
lyskov, line 30: should probably be: `if ( not file1 or not file2)`, see: https://stackoverflow.com/a/24097622/7243597

test/diffbinder.cpp
```diff
@@ -0,0 +1,67 @@
...
+int main(int argc, char** argv)
+{
+    if (argc!=3&&argc!=4) {
+        std::cout<<"Wrong number of arguments"<<std::endl;
+        return EXIT_FAILURE;
+    }
+    int status=COMPARE_ASCII_FILES(argv[1],argv[2]);
+    if (argc==4)  { if (strcmp(argv[3],"--always-success")==0)  return EXIT_SUCCESS; }
+    return status;
+}
```
lyskov, line 67: @andriish i added only some comments for this file and will add more when/if we decide if using custom-made utility is a sound approach. Here is my thoughts on this:

-  I do agree with your point that using just plain `diff` might not be optimal due to potential use on non GNU systems
- using custom C++ code might have benefits if Python is not present on target system
- however since _we do need python to run tests_, - why should not we just use Python scripts instead? 
- using Python script will allow to unify self-test suite with these tests so same comparison logic is used

Could you please argue why using C++ instead of Python will be a better choice? Thanks,

CMakeLists.txt
andriish: Done.

cmake/Modules/CMakeFindFrameworks.cmake
```diff
@@ -0,0 +1,32 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
```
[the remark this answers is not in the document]
andriish: >  Can we enable 'real' comparison phase (ie one that will trigger error) on compatible systems?
Yes, but not immediately.  My understanding is that  in some cases binder relies on the standard behaviour of clang tools, which is slightly different from version to version. Therefore, for many tests there is at least one version of clang/llvm that would fail the comparison test. The solution is obvious: set the behaviour explicitly. But that would require some more time. Therefore my suggestion would be to look case-by-case on the differences and either ignore them, or, if you find the important make sure different versions of clang interpret things in the same way. 


> can we detect that we compile under CentOS-7 with Clang-3.4 and if so use comparison result as pass/fail criteria? 
Yes. CentOS7/clang3.4 support is implemented in another branch.

> Thoughts?
1) If I would decide on the infrastructure in our field, I would switch to CentOS8 some years before -> use the default version of clang on that system for comparison. 
2) In case of implementation of  "set behaviour explicitly "from the answer to the first question, it should not matter.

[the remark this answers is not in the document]
andriish: > This file and others in cmake/Modules: what is the source of these files? Are they coming from Cmake tree? 
Yes.  https://github.com/Kitware/CMake/tree/master/Modules      (but not  the master)

>Can we some where note that? Possibly readme i this dir?
Typically people just copy-paste them "as is" because the files contain the references. But why not. Maybe in the general readme.

test/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+  if (${MAJ} STREQUAL 0 )
+    if (NOT TARGET diffbinder)
+      add_executable( diffbinder diffbinder.cpp) 
+    endif()
+    add_test( NAME ${testname}_diff
+#--always-success option forces the diffbinder to return success regardles of the results of comparison. 
+      COMMAND diffbinder ${testnamenodot}.cpp ${testname}.ref  --always-success
+       WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}")
+    SET_TESTS_PROPERTIES(${testname}_diff PROPERTIES DEPENDS  binder)
+    SET_TESTS_PROPERTIES(${testname}_diff PROPERTIES DEPENDS  ${testname}_foo )
```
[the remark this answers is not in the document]
andriish: Fixed.

test/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+    target_include_directories( ${testnamenodot}${vers} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}  ${pybind11_INCLUDE_DIR})
+
+    if(BINDER_USE_PYTHON_IN_TEST)
+      add_test( NAME ${testname}_python${vers} 
+      COMMAND ${Python_EXECUTABLE} -c "import ${testnamenodot}"
+        WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/py${vers}/${CMAKE_INSTALL_LIBDIR}/$<0:>)
+    else()
+      add_test( NAME ${testname}_python${vers} 
+      COMMAND echo "Import for ${testnamenodot} is disabled"
+        WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/py${vers}/${CMAKE_INSTALL_LIBDIR}/$<0:>)
```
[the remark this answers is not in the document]
andriish: Fixed.

test/diffbinder.cpp
andriish: Done.

test/diffbinder.cpp
andriish: Done.

test/diffbinder.cpp
andriish: OK. That is simple. Once everything is working, the simplest set of tests to run is just the diff test "-DBINDER_TEST_PYTHON_VERSIONS=0".   These test don't need python at all. One can also run them on the systems w/o pybind11 headers.

CMakeLists.txt
andriish: >  will tests run on all available pythons

Yes. All the pythons according to the specification.

3 -> the first found python of version 3.*
3.X -> the first found python of version 3.X.*
3.X.Y -> the first python of version 3.X.Y

Same for the version 2.

If there are two installations of the same python, one can prioritize one installation via the standard cmake variables.
( I don't remember them now, but these are described in cmake docs. )

>will remain installed on most systems probably forever. 

Yes. So as for me the defaults -DBINDER_TEST_PYTHON_VERSIONS=0,2,3   are reasonable.
Note that the order does not matter. 0,2,3 is the same as 3,2,0. Likely only the order of compilation will differ.

test/diffbinder.cpp
andriish: Done.

test/diffbinder.cpp
andriish: Ok, ok. C++11. Done.

test/CMakeLists.txt
```diff
@@ -0,0 +1,133 @@
...
+         T15.inner_class
+         T17.anonymous
+         T20.template
+         T30.include
+         T31.include_for_class
+         T32.call_guard
+         T40.stl
+         T42.stl.names
+         T50.namespace_binder
+         )
```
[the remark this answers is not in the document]
andriish: Yes. One can do something like "create tests for all files that match pattern".  Should be a line or two in cmake language.
But it would be better to have it in a separate MR.

test/diffbinder.cpp
andriish: Done.

andriish:
Hi @lyskov ,

here is a list of differences between the references and the generated binders;

```
Fedora 26
220/220-th strings are not equal /github/workspace/test/T10_inheritance.cpp /github/workspace/test/T10.inheritance.ref
->	}<-
->		cl.def( pybind11::init( [](B const &o){ return new B(o); } ) );<-

Fedora 26/27
23/23-th strings are not equal /github/workspace/test/T01_enum.cpp /github/workspace/test/T01.enum.ref
->		.value("E1_V0", E1::E1_V0)<-
->		.value("E1_V0", E1_V0)<-
51/51-th strings are not equal /github/workspace/test/T07_class.cpp /github/workspace/test/T07.class.ref
->			.value("E1V", Enums::E1::E1V)<-
->			.value("E1V", Enums::E1V)<-
65/63-th strings are not equal /github/workspace/test/T11_override.cpp /github/workspace/test/T11.override.ref
->		pybind11::function overload = pybind11::get_overload(static_cast<const Base *>(this), "foo_sring");<-
->	class std::basic_string<char> foo_sring() override { <-
34/44-th strings are not equal /github/workspace/test/T42_stl_names.cpp /github/workspace/test/T42.stl.names.ref
->}<-
->	{ // std::set file:bits/stl_set.h line:90<-

Fedora 28/29/31
65/63-th strings are not equal /github/workspace/test/T11_override.cpp /github/workspace/test/T11.override.ref
->		pybind11::function overload = pybind11::get_overload(static_cast<const Base *>(this), "foo_sring");<-
->	class std::basic_string<char> foo_sring() override { <-
35/44-th strings are not equal /github/workspace/test/T42_stl_names.cpp /github/workspace/test/T42.stl.names.ref
->}<-
->	{ // std::set file:bits/stl_set.h line:90<-

Fedora 32/33
124/124-th strings are not equal /github/workspace/test/T02_function.cpp /github/workspace/test/T02.function.ref
->	M("").def("foo_f1", (void (*)(const class std::function<bool (const class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char> > &)> &)) &foo_f1, "C++: foo_f1(const class std::function<bool (const class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char> > &)> &) --> void", pybind11::arg("f"));<-
->	M("").def("foo_f1", (void (*)(const class std::function<bool (const std::string &)> &)) &foo_f1, "C++: foo_f1(const class std::function<bool (const std::string &)> &) --> void", pybind11::arg("f"));<-
65/63-th strings are not equal /github/workspace/test/T11_override.cpp /github/workspace/test/T11.override.ref
->		pybind11::function overload = pybind11::get_overload(static_cast<const Base *>(this), "foo_sring");<-
->	class std::basic_string<char> foo_sring() override { <-
35/44-th strings are not equal /github/workspace/test/T42_stl_names.cpp /github/workspace/test/T42.stl.names.ref
->}<-
->	{ // std::set file:bits/stl_set.h line:90<-
```
It looks that some differences are because of the absence/presence of includes in the ``.hpp`` files, while the others are because of differences between the stl versions ( this leads to different numbers of lines ). These differences can be eliminated adding the includes to the tests and/or adding extra directives to the configs.
The only "dangerous" difference, as for me, is the diffence in the namespaces: like this.
```
->		.value("E1_V0", E1::E1_V0)<-
->		.value("E1_V0", E1_V0)<-
```

Best regards,

Andrii

lyskov:
> Multiple python versions can be and should be used in parallel. The versions are set as a comma-separated list passed by the BINDER_TEST_PYTHON_VERSIONS option. The default list is 0,2,3.
The Python version "0" corresponds to a "plain diff" of the output vs reference. The versions of python2/Python3 will be the first versions found by cmake. For a specific versions one can use e.g.
cmake .... -DBINDER_TEST_PYTHON_VERSIONS=0,2.7.15,3.8.0,3.7.0 . The generated codes will be compiled and loaded using the corresponding interpreter.

> With an option -DBINDER_MOCK_TEST one can mock the code generation by binder. In this case the reference codes will be used in the python tests.

@andriish could you please add ~this into documentation? I think creating a new `testing.rst` and putting it there would be best.

CMakeLists.txt
lyskov: 👍 Great, - this sounds good to me!

cmake/Modules/CMakeFindFrameworks.cmake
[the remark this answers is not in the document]
lyskov: > If I would decide on the infrastructure in our field, I would switch to CentOS8 some years before -> use the default version of clang on that system for comparison.

-- i hear you! But since HPC wold will be stuck with CentOS-7 for quite a while i am not sure if moving to CentOS-8 will make things easier for average user.

test/CMakeLists.txt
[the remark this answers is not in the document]
lyskov: ok, deferring this to separate PR

test/diffbinder.cpp
lyskov: 👍

test/diffbinder.cpp
lyskov: 👍

test/diffbinder.cpp
lyskov: 👍

test/diffbinder.cpp
lyskov: 👍

lyskov:
> It looks that some differences are because of the absence/presence of includes in the `.hpp` files, while the others are because of differences between the stl versions ( this leads to different numbers of lines ). These differences can be eliminated adding the includes to the tests and/or adding extra directives to the configs.
> The only "dangerous" difference, as for me, is the diffence in the namespaces: like this.
> 
> ```
> ->		.value("E1_V0", E1::E1_V0)<-
> ->		.value("E1_V0", E1_V0)<-
> ```
 
-- could it be that your branch have not latest master? I think this particular diff was recently fixed in master: https://github.com/RosettaCommons/binder/commit/c8a86783187093698d516b5804b2e053655325cd#diff-19c562e123193ce3a3e61a97992d43f2

andriish:
>  add ~this into documentation

Done.

andriish:
> could it be that your branch have not latest master?
That is the version from master.   The point is that on some systems/llv the result is different from the reference, but on some others it is the same. This tet "fails" only on Fedora26/27. Works fine for others. To some extent, the "fix" has fixed the result for Fedora 28-33 and broke the result for 26-27.

  Please note that I don't blame "old LLVM" because the reason could be something else. So far I don't know it.

test/diffbinder.cpp
andriish: > "finding Python interpreter"
Some systems have python, some have python2, some python3.  
To have a python interpreter in a build system (cmake or autotools) in a portable way one has to find it and not expect 
that python/python2/python3/python.exe is in PATH.
That stuff is always ugly.

>  the regex version from GCC-4.8 shipped with CentOS-7
Yes, I know. Very sad story. But hopefully, in the end,  there will be no need for regexps (i.e. the whole function 'skip' will not be needed) and the outputs will be always the same.

>command-line-options parsing
That can be added, there are standard superportable solutions. But diffbinder is an internal utility.

> C++ solution is superior or because current implementation is already written?
As for me, it is not superior because the task is trivial. For the same reason, one needs very little time to implement it in any language, so "already exist" is not a strong argument. 
But the C++ implementation is much more portable and does not require any scripting in cmake.

Practically: one can add an alternative 'diffbinder.py' and remove the c++ version in another MR.

An esoteric option:  write the diff as cmake function. Regexps and portability included.

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (s1.find("pybind11::handle cl_type = cl;")!=std::string::npos) return 6;  //This could be optimized in binder
+    if (s1.find("__cxx11")!=std::string::npos) return 7;                         //This looks like implementation dependent
+    if (s1.find("assign")!=std::string::npos) return 8;                          // The assing is binded differently for some reasons
+    if (s1.find("#include")!=std::string::npos) return 9;                        // Includes could differ
+    if (s1.find("pybind11::gil_scoped_acquire gil;")!=std::string::npos) return 10; //This could be optimized in binder
+    return 0;
+}
+std::string strip(std::string const & s1)
+{
+    if (s1.length()==0) return s1;
```
lyskov, line 32: optional (feel free to ignore this): `if( s1.empty() )...`

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+    size_t comment=s1.find_first_of("//");
+    return s1.substr(0,comment);
+}
+int compare_text_files(std::string const & f1,const std::string & f2)
+{
+    std::fstream file1(f1), file2(f2);
+    std::string string1, string2;
+    int j1,j2;
+    j1 = 0;
+    j2 = 0;
```
lyskov, line 42: optional: `int j1=0, j2=0;`

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+            j1++;
+            if (!std::getline(file1,string1)) break;
+            if (skip(string1)==0) break;
+        }
+        for (;;) {
+            j2++;
+            if (!std::getline(file2,string2)) break;
+            if (skip(string2)==0) break;
+        }
+        if(strip(string1).compare(strip(string2)) != 0)
```
lyskov, line 61: optional: why not just `if( strip(string1) != strip(string2) )`?

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+int main(int argc, char** argv)
+{
+    if (argc!=3&&argc!=4) {
+        std::cout<<"Wrong number of arguments"<<std::endl;
+        return EXIT_FAILURE;
+    }
+    int status=compare_text_files(argv[1],argv[2]);
+    if (argc==4)  { if (strcmp(argv[3],"--always-success")==0)  return EXIT_SUCCESS; }
+    return status;
```
lyskov, line 79: What if last argument is not `--always-success` but `--i-do-not-know-what-options-is-accepted`? I think it would be best to print help message (below) and return non-zero exit-code in such case.

Also, could you please add minimal description to what this utility do and minimal help message (with possible options) when wrong number of arguments is used?

test/diffbinder.cpp
lyskov: ok, lets accept this now, and we can revisit this later when/if expanding of this utility became an issue.

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (s1.find("Source list file")!=std::string::npos) return 3;       //Same as the list of source files
+    if (s1.find("Modules list file")!=std::string::npos) return 4;      //Same as the list of modules
+    if (s1.find("File:")!=std::string::npos) return 5;                  //Same as the full file paths.
+    if (s1.find("pybind11::handle cl_type = cl;")!=std::string::npos) return 6;  //This could be optimized in binder
+    if (s1.find("__cxx11")!=std::string::npos) return 7;                         //This looks like implementation dependent
+    if (s1.find("assign")!=std::string::npos) return 8;                          // The assing is binded differently for some reasons
+    if (s1.find("#include")!=std::string::npos) return 9;                        // Includes could differ
+    if (s1.find("pybind11::gil_scoped_acquire gil;")!=std::string::npos) return 10; //This could be optimized in binder
+    return 0;
+}
```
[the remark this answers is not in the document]
andriish: > Are you planning to somehow use return code?
This could be used to control the tolerance. After some ordering it could be like 1-5 is OK, but anything else is not.
Enum is an option, yes, but so far it is not clear how many of these checks will survive.

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+//
+// All rights reserved. Use of this source code is governed by a
+// MIT license that can be found in the LICENSE file.
+#include <iostream>
+#include <fstream>
+#include <stdio.h>
+#include <vector>
+#include <string.h>
+
+int skip(std::string const & s1)
```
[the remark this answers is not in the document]
andriish: Ah, ok. I will add proper comments in the next commit to all the functions in the code.

test/diffbinder.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (s1.find("Modules list file")!=std::string::npos) return 4;      //Same as the list of modules
+    if (s1.find("File:")!=std::string::npos) return 5;                  //Same as the full file paths.
+    if (s1.find("pybind11::handle cl_type = cl;")!=std::string::npos) return 6;  //This could be optimized in binder
+    if (s1.find("__cxx11")!=std::string::npos) return 7;                         //This looks like implementation dependent
+    if (s1.find("assign")!=std::string::npos) return 8;                          // The assing is binded differently for some reasons
+    if (s1.find("#include")!=std::string::npos) return 9;                        // Includes could differ
+    if (s1.find("pybind11::gil_scoped_acquire gil;")!=std::string::npos) return 10; //This could be optimized in binder
+    return 0;
+}
+std::string strip(std::string const & s1)
```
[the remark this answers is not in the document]
andriish: Same as above.

test/diffbinder.cpp
andriish: OK. I see. Will add docs there.

test/diffbinder.cpp
andriish: OK.

test/diffbinder.cpp
andriish: OK.

test/diffbinder.cpp
andriish: OK.

test/diffbinder.cpp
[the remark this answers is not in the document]
lyskov: got it, thank you for clarifying @andriish !

andriish:
Hi @lyskov  ,
It makes sense to merge it. At least 
One improvement (automatic test discovery) will follow, but in general it would be interesting to eliminate the differences in the tests. And it wouldbe easier to do once this MR ist merged

Best regards,
Andrii

lyskov:
Thank you for putting this together @andriish !

PR #122, opened 2020-08-07
Enable optional static compilation of Binder

### LINKED ISSUE #112: Difficulty statically compiling binder on CentOS 8
Hi! I've been trying to statically compile `binder` on CentOS 8 and have been having difficulty. Have you attempted this before? If so, what difficulty did you encounter in doing so?

I would much appreciate any feedback you have, so I can decide whether it makes sense to continue this effort.

---

Here's some more information on my current progress. I've modified `CMakeLists.txt` per [this stackoverflow answer](https://stackoverflow.com/a/24671474):
```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.0)
 PROJECT(binder CXX C)
 SET(VERSION 1.0.0)
+SET(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+SET(BUILD_SHARED_LIBS OFF)
+SET(CMAKE_EXE_LINKER_FLAGS "-static")
 #This actually will work even with C++11
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_VERBOSE_MAKEFILE ON)
```

But I've run into a problem. The shared library version of libllvm is available in its entirety as `libLLVM.so`, as reflected by this CMake snippet which finds that library:

https://github.com/RosettaCommons/binder/blob/022cd0d5495628e5b50d7ffb6eb0abb6ccd543e3/source/CMakeLists.txt#L37-L42

The static library version of libllvm, on the other hand, is distributed across many static archives:

```bash
$ ls -1 /usr/lib64/libLLVM*.a
/usr/lib64/libLLVMAArch64AsmParser.a
/usr/lib64/libLLVMAArch64CodeGen.a
/usr/lib64/libLLVMAArch64Desc.a
/usr/lib64/libLLVMAArch64Disassembler.a
/usr/lib64/libLLVMAArch64Info.a
/usr/lib64/libLLVMAArch64Utils.a
...
(etc.)
```

Because I don't know which static archive(s) Binder depends on, I'm not sure how to modify the `find_library` call referenced above. This is the farthest I've gotten.

ultra4021:
Hi @lyskov! You may notice that this PR lacks much documentation. I wanted to make sure that this approach is the best way to do things before I documented it. The rest of the PR is as follows.

---

This pull request allows users to statically compile `binder` for easier deployment. This depends on deech's [libclang-static-build](https://github.com/deech/libclang-static-build), which builds a static archive that bundles `libclang-10.0.0` and `libllvm-10.0.0` (the LLVM project does not officially provide a static variant of libclang, AFAICT).

To statically compile binder, first build `libclang-static-build`, which has *excellent* instructions. Then do this in a shell:

```bash
mkdir build
cd build
cmake .. -DSTATIC=on \
    -DLLVMCONFIG="${LIBCLANG_STATIC_BUILD_DIR}/build/_deps/libclang_prebuilt-src/bin/llvm-config" \
    -DLLVM_LIBRARY_DIR="${LIBCLANG_STATIC_BUILD_DIR}/lib"
make -j4
```

where `${LIBCLANG_STATIC_BUILD_DIR}` is set to the path of the `libclang-static-build` repository.

Closes #112.

Tested on CentOS 8 x86_64.


andriish:
Hi @ultra4021 ,
The updated CMakeLists.txt  still contains lib_llvm_so_path.

Best regards,

Andrii

ultra4021:
Thanks Andrii, just fixed that and made the error messages better when `libclang_static_bundled` (and `libllvm.so`) cannot be found. See the commit messages for more information

andriish:
Hi @ultra4021 ,

is the ``libclang-static-build`` a build from the LLVM/Clang developers?
I'm asking because I'm slightly confused with the idea of the static build of LLVM and not actually sure if it is useful in this context.

As far as I understood you are trying to create a single executable for binder that can be copied from one system to another without a need for deployment of the LLVM/Clang ( I like that idea). But there is a catch:  even in this approach, one would need to have the standard C++ headers on the system where binder will be running.  If my understanding is correct, to be sure things work properly these should be libc++ headers of compatible version. This means that in addition to binder executable itself, one should also deploy the headers and some other files and practically there is no advantage of the deployment of ``statically'' build binder vs. dynamically build binder.  
 Maybe I'm wrong, I haven't tried to do this build by myself. But have you tried to run the ``statically`` build executable on some other system w/o clang/llvm installation?


Best regards,

Andrii

ultra4021:
> is the `libclang-static-build` a build from the LLVM/Clang developers?

`lib-clang-static-build` is not from LLVM/Clang developers.

> As far as I understood you are trying to create a single executable for binder that can be copied from one system to another without a need for deployment of the LLVM/Clang ( I like that idea). But there is a catch: even in this approach, one would need to have the standard C++ headers on the system where binder will be running. If my understanding is correct, to be sure things work properly these should be libc++ headers of compatible version.

Ah, I did not consider that...

> But have you tried to run the `statically`` build executable on some other system w/o clang/llvm installation?

I have, and oddly enough everything works fine, perhaps because there is no breaking difference between the libc++ on CentOS 8 (where I compiled Binder) and RHEL 7 (where I deployed Binder).

andriish:
Hi @ultra4021 ,

> I have, and oddly enough everything works fine, perhaps because there is no incompatibilities between the libc++ on CentOS 8 (where I compiled Binder) and RHEL 7 (where I deployed Binder)?

ok, that is interesting to know.

If you actually want to deploy binder on the standard CentOS7 but stick to the official releases of the other software, there are two ways: either use the ``llvm`` of version 7 from the ``llvm-toolset`` repos for the CentOS7, or patch binder to make it work with the default CentOS7 llvm 3.4. The first option is described in the documentation of binder (recommended!), the second option I've tried to implement here: https://github.com/andriish/binder/tree/CentOS7.

Best regards,

Andrii

ultra4021:
Andrii, for our needs it's easier to avoid shared libraries (long story), but I appreciate you looking into CentOS-7.

lyskov:
@ultra4021 looks good to me! As @andriish mentioned in order for this approach to work C++ headers on build and deploy system should be somewhat compatible though some deviation will likely to not cause issues. So i think we can merge this but let clearly state in documentation that (a) this is somewhat experimental feature and (b) explain C++ headers caveat. Could you please add relevant part to the docs? I think this extra docs should be put somewhere in https://github.com/RosettaCommons/binder/blob/master/documentation/install.rst Thanks,

lyskov:
Also, - glad to hear that it possible to link Binder statically! @ultra4021 have you tried running such build on fairly different system? Like say CentOS-7 vs latest Debian/Ubuntu? I am curios how far we can stretch this until deviation in C++ headers will became an issue.


> I have, and oddly enough everything works fine, perhaps because there is no incompatibilities between the libc++ on CentOS 8 (where I compiled Binder) and RHEL 7 (where I deployed Binder)?

-- how did you test this? Was it just Binder self-test run or something more elaborate? I am asking because discrepancies in C++ headers will start to play when Binder tries to generate a lot of bindings for STL code. This btw could also be mentioned in related docs: as long as user tries to bind non-system code results of such runs might be good to use.

andriish:
Hi @lyskov ,

1) concerning this MR: another minor thing is that version of LLVM for this MR is hardcoded. As of now the option is designed to work only for LLVM 10  just because the github user "deech" has  put online the scripts only for LLVM 10.  

2) 
> I am curios how far we can stretch this until deviation in C++ headers will became an issue.

I've looked at this topic a bit before.
Within LLVM toolchain  I would expect a compatibility table from 
http://libcxx.llvm.org/docs/  , but everything what I've found  for the current dev version is 
```
The following minimum compiler versions are required:
    Clang 4.0 and above
    GCC 5.0 and above.
```
In a wider context,  there is an option to "fake" the behavior of different compilers by LLVM/Clang and assure almost complete portability. Something that is implemented here:  https://github.com/CastXML/CastXML/blob/master/doc/manual/castxml.1.rst 
If implemented in binder, this feature could allow to "cross-generate" bindings, i.e. generate MSVC bindings on Linux.
But I don't know if that is what you would be interested in. At least from the point of work over profit ratio.

Best regards,

Andrii

ultra4021:
> @ultra4021 looks good to me! As @andriish mentioned in order for this approach to work C++ headers on build and deploy system should be somewhat compatible though some deviation will likely to not cause issues. So i think we can merge this but let clearly state in documentation that (a) this is somewhat experimental feature and (b) explain C++ headers caveat. Could you please add relevant part to the docs? I think this extra docs should be put somewhere in https://github.com/RosettaCommons/binder/blob/master/documentation/install.rst Thanks,

@lyskov Just added documentation 👍 

> Also, - glad to hear that it possible to link Binder statically! @ultra4021 have you tried running such build on fairly different system? Like say CentOS-7 vs latest Debian/Ubuntu? I am curios how far we can stretch this until deviation in C++ headers will became an issue.

@lyskov I have not, only on CentOS 7, CentOS 8, and RHEL 7. Unfortunately I will not be able to do additional testing anytime soon due to crunch time at work 😅 .


> -- how did you test this? Was it just Binder self-test run or something more elaborate? I am asking because discrepancies in C++ headers will start to play when Binder tries to generate a lot of bindings for STL code. This btw could also be mentioned in related docs: as long as user tries to bind non-system code results of such runs might be good to use.

@lyskov Good question! It was a test run on a large (~50,000 SLOC) C++ codebase. And that's a good point about STL code; we didn't run into issues there because we don't bind much if any STL code.

ultra4021:
> concerning this MR: another minor thing is that version of LLVM for this MR is hardcoded. As of now the option is designed to work only for LLVM 10 just because the github user "deech" has put online the scripts only for LLVM 10.

@andriish That's a good point. I added a comment to `source/CMakeLists.txt` clarifying why static compilation only works for LLVM 10.

Also I just pushed documentation and rebased onto master.

lyskov:
> In a wider context, there is an option to "fake" the behavior of different compilers by LLVM/Clang and assure almost complete portability. Something that is implemented here: https://github.com/CastXML/CastXML/blob/master/doc/manual/castxml.1.rst
> If implemented in binder, this feature could allow to "cross-generate" bindings, i.e. generate MSVC bindings on Linux.
> But I don't know if that is what you would be interested in. At least from the point of work over profit ratio.

-- I am not sure if that will be possible without shipping Binder with STL sources for different platforms/compiler-versions. Maybe CastXML docs meant something different?

lyskov:
> * concerning this MR: another minor thing is that version of LLVM for this MR is hardcoded. As of now the option is designed to work only for LLVM 10  just because the github user "deech" has  put online the scripts only for LLVM 10.

-- good point @andriish ! Looks like @ultra4021 added documentation stating this so i think it is ok for now (since it is experimental) But please let me know if you think there is a point of differing this merge because of this.

> And that's a good point about STL code; we didn't run into issues there because we don't bind much if any STL code.

-- thank you for letting me know @ultra4021 ! I will see if i can try found time to use static build with our project, due to its size (~3M SLOC) we have loads of STL bindings, so that might give a good representation of how portable this solution is.

ultra4021:
Thanks for the review @lyskov! I'd be glad to look into CI, but it will have to wait until next week due to work.

andriish:
Hi @lyskov,

> -- I am not sure if that will be possible without shipping Binder with STL sources for different platforms/compiler-versions.
That was the point: to have an option to fit to the already *installed* STL. Shipping some ripped-off copy of STL would be hard to implement and very messy.

> Maybe CastXML docs meant something different?
Very likely. My simple understanding was that CastXML with these options and corresponding STL implementations would be able to work reasonably, i.e. produce proper AST in xml.  But one would need some time and work to actually understand what is happening in reality.

Best regards,

Andrii

andriish:
Hi @lyskov, @ultra4021 

 > would you be interested to add a CI test for this? (CC'ing @andriish who might have thoughts on this).

Hm. I'm looking at the libraries I have now in CentOS7 and see that there are the static libraries for everything.
I will try to figure out if one actually needs that non-official repository of user deech.
Maybe it is possible to have a static build  and CI  with the official distribution. 


Best regards,

Andrii

andriish:
Hi @lyskov , @ultra4021 ,

it turns out the static build (*almost ) is possible if the clang and llvm were build statically. 
I've implemented this option in a separate branch and it works, but 

-- Very few distributions provide clang and llvm with statically build libraries.  CentOS7  (LLVM 3.4+) does.   CentOS8 and Fedora26+ ( LLVM4+) dont.  Means there are no precompiled packages for LLVM/CLANG versions supported by binder. One should have a look at some other distros. But the chances are high that no distro provides them officially.  So 
   --  It could be tricky to organize a CI for static builds.
   -- The static compilation of binder for llvm/clang4+ would look like a compilation of LLVM  + compilation of binder.  Same amount of time as a compilation inside of LLVM tree.


 It  any case it makes sense to implement the CI for special cases only after the #121  is merged. Otherwise it would be too much of mess simultaneously.

Best regards,

Andrii


* only the standard libraries are linked dynamically, i.e.

```
+ ldd source/binder
	linux-vdso.so.1 =>  (0x00007ffcc07e1000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fb437c71000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fb437a55000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007fb437851000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fb43754f000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fb437339000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fb436f6b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb437f78000)

```

ultra4021:
@andriish Thanks a lot for looking into static builds further. And I *completely* forgot to mention this, but my internship is ending this week so I won't be able to work in this further---apologies for the surprise.

Regardless, I've just rebased this PR on top of master and everything seems to be working. Additionally, me and my coworker have been using a static version of Binder frequently the past week (on the aforementioned large codebase) and have encountered no issues not already present in the shared-lib version of Binder.

Thanks again @lyskov and @andriish for your help so far in reviewing!

PR #123, opened 2020-08-08
Extension of documentation: how to point binder to the desired headers

andriish:
Hi @lyskov ,

this MR contains an update of documentation for usage of binder compiled with standalone LLVm in the case if clang is not a system compiler.

This is motivated by #119 

Best regards,
Andrii


andriish:
Also implements suggestion about headers in #122

Issue #57, opened 2018-11-29
Binder appears to write main cpp file before checking if output directory exists


fake-name:
This is a weird edge-case that might be a confluence of a number of things.

Basically, I have a project built as follows

```
#!/bin/bash

rm -rfv /media/sf_Code/scfft/libraries/sndlib_py/generated_binding/
sync

/home/durr/binder/build/llvm-4.0.0/build_4.0.0.linux.desktest.release/bin/binder \
	-v \
	--root-module sndlib_py                                                       \
	--prefix /media/sf_Code/scfft/libraries/sndlib_py/generated_binding/         \
	--config binder_config.config                                                \
	--bind "my_namespace"                                                               \
	binder_includes.hpp                                                          \
	--                                                                           \
	 -std=c++11                                                                  \
	-I/media/sf_Code/scfft                                                       \
	-DNDEBUG
```

If I have the `rm` command remove everything *including* the `generated_binding` directory, the generator runs without error and produces the expected output for everything, but `sndlib_py.cpp` never gets written. However, if I change the `rm` command to not delete the directory:

```
rm -rfv /media/sf_Code/scfft/libraries/sndlib_py/generated_binding/*
```
 `sndlib_py.cpp`  gets successfully written.

There's a bunch of confounding factors, but it appears that the writing of at least *some* of the output files is not properly checked for success. Additionally, verbose mode doesn't ever mention the writing of  `sndlib_py.cpp`, though it does mention all the other generated `*.cpp` files.

I'm working in a virtualbox VM which is mounting a host directory, so there could be shared folder oddness going on, but the issue seems to be quite reproducible.

fake-name:
Ok, I moved into a plain folder on the local drive, and can confirm this exists without all the network/shared folder oddments.

Basically, if the path specified by `--prefix` doesn't exist, it is created, but `{root-module}.cpp` is never created.

lyskov:
@fake-name thank you for reporting this! The prefix directory is indeed should be present before run so such failure is expected, however we should indeed report an error in such cases. I will looks this up.

fake-name:
The prefix directory is created during the run, but I think there's a race condition, or a mis-ordering so it's created after the attempt to write the module root. The issue is repeatable, at least for me.

HFTrader:
I struggle with this as well until I realized the problem. It's interesting this issue has been hanging for a couple of years, seems easy enough to me to solve. So either:
1. Check if the prefix folder exists and bail out if not; or
[...]

PR #124, opened 2020-08-11
Checks if the prefix directory exist before writing the main file. Solves #57 

HFTrader:


source/util.cpp
```diff
@@ -99,20 +100,35 @@ string indent(string const &code, string const &indentation)
...
+std::string join( std::vector<std::string>& splits, std::string const& sep ) {
+	if ( !splits.empty() ) {
+		std::ostringstream ss;
+		ss << splits[0];
+		for ( unsigned j=1; j<splits.size(); ++j ) {
+			ss << sep << splits[j];
+		}
+  	return ss.str();
+	}
+	return std::string();
```
lyskov, line 112: I think we can simplify this a bit if we use STL algorithms, how about something like:
```C++
std::ostringstream o;
std::copy(splits.begin(), splits.end(), std::ostream_iterator<std::string>(o, sep));
std::string r = o.str();
r.resize(r.size() - sep.size());
return r;
```

source/util.cpp
```diff
@@ -99,20 +100,35 @@ string indent(string const &code, string const &indentation)
 	return r;
 }
 
+std::string join( std::vector<std::string>& splits, std::string const& sep ) {
```
lyskov, line 103: `splits` should be passed by `const &`

source/util.cpp
lyskov, line 112: I think this could be simplified a bit, how about:
```C++
std::string join( std::vector<std::string> const &splits, std::string const& sep )
{
	std::string r;
	for(auto const & s : splits) { r += s; r+= sep; }
	if( r.size() ) r.resize(r.size() - sep.size());
	return r;
}
```

source/util.cpp
```diff
@@ -99,20 +100,35 @@ string indent(string const &code, string const &indentation)
...
+
+std::string dirname( std::string const& fullpath ) {
+	vector<string> dirs = split(fullpath, "/");
+	dirs.pop_back();
+	return join( dirs, "/" );
+}
+
+void create_directories( std::string const& dirname ) {
+	string command_line = "mkdir -p " + dirname;
+  system( command_line.c_str() );
```
lyskov, line 123: indentation here is needs to be fixed

source/util.hpp
```diff
@@ -48,7 +48,14 @@ std::string indent(std::string const &code, std::string const &indentation);
 /// Try to read exisitng file and if content does not match to code - write a new version. Also create nested dirs starting from prefix if nessesary.
 void update_source_file(std::string const &prefix, std::string const &file_name, std::string const &code);
 
+/// Joins the vector into a string using the given separator
+std::string join( std::vector<std::string>& splits, std::string const& sep );
```
lyskov, line 52: again, `splits` should be taken by `const &`. Also i recommend to rename it something more descriptive, like `parts`

source/util.cpp
HFTrader: In general, std::ostringstream is optimized for appends, which std::string in itself is not. What you are proposing is way more expensive and would have a big impact in large batches in general. In this case, not very much.
As far as simplifying, as I read it looks more complex but that's just me.

source/util.hpp
HFTrader: I realized that after I checked in.

source/util.cpp
HFTrader: One thing I was not sure - did you leave prefix open such that you could actually add that as file name prefix? Eg if
prefix = /home/abc/foo
root_module=bar
Do you want the files to be created as 
/home/abc/foo/bar.cpp  or
/home/abc/foobar.cpp ?

andriish:
Hi all, 
shouldn't be this

https://github.com/RosettaCommons/binder/compare/master...andriish:issue57?expand=1

 enough?

``mkdir()`` unlike ``system("mkdir -p")`` is POSIX-portable, way faster than ``system()`` and actually takes care about the separators so 
```
mkdir(".///xxx/////", ACCESSPERMS);
mkdir(".///xxx////yyyyy//", ACCESSPERMS);
```
is fine.

Best regards,
Andrii

source/util.cpp
lyskov: > In general, std::ostringstream is optimized for appends

-- Thank you for info @HFTrader ! Interesting, i was not aware of this. Could you please point me where this is specified? Because i was under impression that `std::ostringstream` is just a wrapper on `std::string` which is used as a buffer. For instance, cppreference.com is explicitly stating it here: https://en.cppreference.com/w/cpp/io/basic_ostringstream . Thanks,

lyskov:
> Hi all,
> shouldn't be this
> 
> https://github.com/RosettaCommons/binder/compare/master...andriish:issue57?expand=1
> 
> enough?
> 
> `mkdir()` unlike `system("mkdir -p")` is POSIX-portable, way faster than `system()` and actually takes care about the separators so
> 
> ```
> mkdir(".///xxx/////", ACCESSPERMS);
> mkdir(".///xxx////yyyyy//", ACCESSPERMS);
> ```
> 
> is fine.
> 
> Best regards,
> Andrii

Oh, i like that solution @andriish - much less code to maintain! Would you be open to replace `fprintf` call with `std::cerr << ...` call and create PR from it? Thanks,

HFTrader:
note that the main issue is a check on context.cpp.  the other issues on split and mkdir are minor.

lyskov:
@HFTrader i updated master so `prefix` is now treated only as directory: https://github.com/RosettaCommons/binder/commit/a1870cb37b2fb1929a3fdf75ce5ba63430fc2417. I also added some error checks. Could you please see if this latest master works for you? Thanks,

PR #125, opened 2020-08-14
Bind Nested Template Classes

hollinsky:
This change binds template classes nested within another class

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -1024,20 +1024,37 @@ std::string ClassBinder::bind_repr(Context &context)
 }
 
 
-string ClassBinder::bind_nested_classes(CXXRecordDecl const *EC, Context &context)
+void ClassBinder::for_public_nested_classes(const std::function<void(clang::CXXRecordDecl *C)>& f) const
 {
-	string c;
-	for(auto d = EC->decls_begin(); d != EC->decls_end(); ++d) {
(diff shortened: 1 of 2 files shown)


source/class.cpp
```diff
@@ -1024,20 +1024,37 @@ std::string ClassBinder::bind_repr(Context &context)
 }
 
 
-string ClassBinder::bind_nested_classes(CXXRecordDecl const *EC, Context &context)
+void ClassBinder::for_public_nested_classes(const std::function<void(clang::CXXRecordDecl *C)>& f) const
```
lyskov, line 1027: Looks like unction could take `const *` instead of just `*`. Also could you please use the same notation as the rest of the code, ie: `void ClassBinder::for_public_nested_classes(std::function<void(clang::CXXRecordDecl const *C)> const & f) const`? Thanks,

source/class.cpp
```diff
@@ -1024,20 +1024,37 @@ std::string ClassBinder::bind_repr(Context &context)
...
 			}
 		}
 	}
+}
+
+
+string ClassBinder::bind_nested_classes(Context &context)
+{
+	string c;
+	for_public_nested_classes([&c, &context, this](CXXRecordDecl *innerC) {
```
lyskov, line 1049: probably should be `CXXRecordDecl const *innerC`

lyskov:
never mind, - will address these comments on my next path. - thank you for putting this together @hollinsky !

PR #126, opened 2020-08-14
Add Includes for Nested Classes

hollinsky:
If an `+include_for_class` directive is present in the config file for a nested class, it should be processed while processing the relevant includes for the parent.

**Note:** This uses the `for_public_nested_classes` function added in #125, so #125  should be merged before this PR.


lyskov:
Thank you for adding this @hollinsky !

PR #130, opened 2020-08-19
Fixing another test to make it work exacly the same on multiple systems

andriish:
To match the reference source one should use 

```
#define  _GLIBCXX_USE_CXX11 0
```
this assures usage of  ``string`` instead of  `` _cxx11::string`` 

``<memory>`` header is not needed.

### DIFF
--- a/test/T11.override.hpp
+++ b/test/T11.override.hpp
@@ -12,7 +12,9 @@
 
 #ifndef _INCLUDED_T10_override_hpp_
 #define _INCLUDED_T10_override_hpp_
-
+/// Please note that some systems provide dual C++ ABI, which might affect the  generated code.
+/// The line below forces the usage of the older implementation of std::string on the systems with GNU libc.
+/// See  https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html for details.
 #define _GLIBCXX_USE_CXX11_ABI 0 
 #include <string>
 
--- a/test/T11.override.hpp
+++ b/test/T11.override.hpp
@@ -15,7 +15,7 @@
 /// Please note that some systems provide dual C++ ABI, which might affect the  generated code.
 /// The line below forces the usage of the older implementation of std::string on the systems with GNU libc.
 /// See  https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html for details.
-#define _GLIBCXX_USE_CXX11_ABI 0 
+/// #define _GLIBCXX_USE_CXX11_ABI 0 
 #include <string>
 
 class Base
--- a/test/T11.override.hpp
+++ b/test/T11.override.hpp
@@ -13,7 +13,7 @@
 #ifndef _INCLUDED_T10_override_hpp_
 #define _INCLUDED_T10_override_hpp_
 
-#include <memory>
+#define _GLIBCXX_USE_CXX11_ABI 0 
 #include <string>
 
 class Base


lyskov:
@andriish could you please clarify: what exactly the `_GLIBCXX_USE_CXX11` define do/control?

andriish:
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

Very simplified for this case: selects between   string =string and string =_cxx11::string.

lyskov:
ok, i just pushed the fix for this, please see https://github.com/RosettaCommons/binder/commit/eb055a33b972a0d4087a5e3e5161eddf4614ad34 for details.

andriish:
Hi @lyskov ,
> do you happen to know which particular Fedora version give you trouble? 
That is in CI. Even now, with the option --always-success the ref and produced files are compared.
The logs of the tests are shown to termenal in the end of CI.

For instance:  https://github.com/RosettaCommons/binder/runs/1014407405?check_suite_focus=true
Go to Fedora 26, click at "Compli" section and in the very end there will be:

```
 Run comparison
65/63-th strings are not equal /github/workspace/test/T11_override.cpp /github/workspace/test/T11.override.ref
   ->		pybind11::function overload = pybind11::get_overload(static_cast<const Base *>(this), "foo_sring");<-
   ->	class std::basic_string<char> foo_sring() override { <-
<end of output>
Test time =   0.00 sec
```

Alternatively, I download the whole archive, unpack it and run something like

```
 for a in $(ls -1 1*.txt); do echo $a; cat $a | grep -A 2 "not equal" ; done
```
For the debug purposes I was adding "cat  test/T11_override.cpp" to the executed script in CI (therefore I don't have the output now in a good form).
Form the temporary files I'm looking at now I see the difference was 
```
	{ // B file:T10.inheritance.hpp line:53
		pybind11::class_<B, std::shared_ptr<B>> cl(M(""), "B", "");
		cl.def( pybind11::init( [](){ return new B(); } ) );
	}
```
vs

```
	{ // B file:T10.inheritance.hpp line:53
		pybind11::class_<B, std::shared_ptr<B>> cl(M(""), "B", "");
		cl.def( pybind11::init( [](B const &o){ return new B(o); } ) );
		cl.def("assign", (class B & (B::*)(const class B &)) &B::operator=, "C++: B::operator=(const class B &) --> class B &", pybind11::return_value_policy::automatic, pybind11::arg(""));
	}
```


Best regards,
Andrii

andriish:
> cat test/T11_override.cpp

Or T10*. It looks it has bee renamed.

lyskov:
I checked the latest diff (https://pipelines.actions.githubusercontent.com/Delybi8grO9inWHr5Q3X0Kh1N8yR7TbD0hG9URLu9GeKbVW2dQ/_apis/pipelines/1/runs/100/signedlogcontent/12?urlExpires=2020-08-24T22%3A27%3A02.0522472Z&urlSigningMethod=HMACV1&urlSignature=bkcS5q%2FVS3VS7FYBIXD3zRUX32o88XYMO1Vyz6jmJos%3D) and it looks like the differences related to `basic_string` is now gone so latest patch has worked.

However diff related to an extra `assign` operator is still there and i do not see how to fix them except to uncomment `_GLIBCXX_USE_CXX11` difine. I will try to do that shortly. Let's see if that helps.

andriish:
Hi @lyskov ,

I don't remember that exactly ( and im not an expert),
 but the assign
Was the important difference between 
two ABIs. Very likely it is unavoidable

Best regards,

Andrii

P.S. A bit unrelated : some last commits 
made the T42 tests 
Be different again. Will have to look.

lyskov:
Looks like it worked, the assign diff is now gone: https://pipelines.actions.githubusercontent.com/Delybi8grO9inWHr5Q3X0Kh1N8yR7TbD0hG9URLu9GeKbVW2dQ/_apis/pipelines/1/runs/101/signedlogcontent/11?urlExpires=2020-08-24T22%3A55%3A03.7171737Z&urlSigningMethod=HMACV1&urlSignature=pv7%2Ft97NH%2FH4PTIMHc84oVFLvzCrIlUzkYYDC0X%2B%2B9s%3D

PR #131, opened 2020-08-19
Fix test 42 for multiple platforms.

andriish:
The test 42 was split into test 43 and test 44.

Test 43 contains just normal classes and test 44 multi*.

Without this split  the order of generation of bindings was different on different platforms. In some cases the multi* classes were written first, in the other the normal classes preceded multi*.

### DIFF
--- a/test/T42.stl.names.multi.config
+++ b/test/T42.stl.names.multi.config
@@ -1,3 +1,0 @@
--function std::deque<int, std::allocator<int> >::deque(unsigned long)
--function std::list<double, std::allocator<double> >::list(unsigned long)
--function std::vector<double, std::allocator<double> >::vector(unsigned long)


test/T44.stl.names.hpp
```diff
@@ -22,20 +23,15 @@
 #include <unordered_set>
 #include <unordered_map>
 
-void foo(std::set<float> ) {}
 void foo(std::list<double> ) {}
 void foo(std::forward_list<float> ) {}
-void foo(std::map<float,int> ) {}
 void foo(std::vector<double> ) {}
 void foo(std::deque<int> ) {}
```
lyskov, line 29: Why not move these test cases as well? That way we can have 42 only for `multi*`. 

If you agree with above then lets also rename 42 to `T42.stl.names.multi.hpp` to make reason for this separation clear. (and it might be a good idea to put some minimal comment into both files explaining and referencing this PR for clarity). 

Thanks,

andriish:
Done.  

If this and #130 are merged, only the problem with namespaces will remain. 
And a weird thing in T10.inheritance.ref on older llvm.

PR #133, opened 2020-08-21
Revert #128

lyskov:
Reverts RosettaCommons/binder#128


lyskov:
Reverting #133 since it broke multiple-files output mode. Mentioning @andriish who is author of original PR.

PR #134, opened 2020-08-25
Test on rawhide

andriish:
Hi @lyskov,
some recent runs with Fedora32/33 were failing as the repositories for these Fedoras contain now newer LLVM ( a prerelease of LLVM11). This MR solves this issue.

In this MR: 
  -  explicitly added <clang/Basic/SourceManager.h> and string(tc->getText())  to make compilation with LLVM11 possible.
  - A bit more advanced way to search for the required headers using the cmake find_path() function.
    The changes include a workaround  for Fedora32, for which the LLVM 10.0.1 headers are located in   ../10.0.0/   directory.
  - Cosmetics.   

Best regards,

Andrii


CMakeLists.txt
```diff
@@ -26,21 +26,33 @@ MESSAGE(STATUS  "binder: BINDER_USE_PYTHON_IN_TEST=${BINDER_USE_PYTHON_IN_TEST}"
...
+    ${LLVM_LIBRARY_DIR}/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
+    ${LLVM_LIBRARY_DIR}/../clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
+    ${LLVM_LIBRARY_DIR}/../../lib/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
+    ${LLVM_LIBRARY_DIR}/../../lib64/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
+    ${CLANG_INCLUDE_DIRS}/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
+    ${LLVM_LIBRARY_DIR}/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
+    ${LLVM_LIBRARY_DIR}/../clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
+    ${LLVM_LIBRARY_DIR}/../../lib/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
+    ${LLVM_LIBRARY_DIR}/../../lib64/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
+    ${CLANG_INCLUDE_DIRS}/clang/${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
```
lyskov, line 53: This has became quite long. Could you please see if it would be possible to refactor this into for-loop iterating over list of `clang`, `../clang`, `../lib/clang/`, `../lib64/clang/` and second loop iterating over list of [`${LLVM_VERSION}`, `${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}`]? Thanks,

CMakeLists.txt
```diff
@@ -78,13 +90,26 @@ else()
...
     list(GET VERSION_LIST 1 LLVM_VERSION_MINOR)
     list(GET VERSION_LIST 2 LLVM_VERSION_PATCH)
+    find_path(LibClang_INCLUDE_DIR stddef.h HINTS
+      ${LLVMINCLUDEDIR}/clang/${LLVMVERSION}/include
+      ${LLVMINCLUDEDIR}/../clang/${LLVMVERSION}/include
+      ${LLVMINCLUDEDIR}/../../lib/clang/${LLVMVERSION}/include
+      ${LLVMINCLUDEDIR}/../../lib64/clang/${LLVMVERSION}/include
+      ${CLANG_INCLUDE_DIRS}/clang/${LLVMVERSION}/include
+      ${CLANG_INCLUDE_DIRS}
+      )
```
lyskov, line 100: again, maybe see if we can iterate over list of `[ clang, ../clang, ../lib/clang/, ../lib64/clang/ ]`

source/util.cpp
```diff
@@ -255,7 +256,7 @@ string generate_comment_for_declaration(clang::NamedDecl const *decl)
 // extract text from hierarchy of comments
 string get_text(comments::Comment const *C, SourceManager const & SM, SourceLocation previous)
 {
-	if( auto tc = dyn_cast<comments::TextComment>(C) ) return tc->getText();
+	if( auto tc = dyn_cast<comments::TextComment>(C) ) return string(tc->getText());
```
lyskov, line 259: also, i am a bit surprised by this change: Could you please double check that this conversion to string is indeed needed? (it should have happen automatically)

CMakeLists.txt
andriish: Hi @lyskov ,
yes, this is ugly. But the uglies part 
is exactly the workaround for the bugged installation when the location of headers is /10.0.0/  despite the version is 10.0.1.

Unfortunately a loop would not make things easier.   A regexp would help, but it is not supported.
The closest thing would be something like
```
    HINTS
    ${LLVM_LIBRARY_DIR}/clang
    ${LLVM_LIBRARY_DIR}/../clang
    ${LLVM_LIBRARY_DIR}/../../lib/clang
    ${LLVM_LIBRARY_DIR}/../../lib64/clang
    ${CLANG_INCLUDE_DIRS}/clang          
    PATH_SUFFIXES
    ${LLVM_VERSION}/include
    ${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}/include
    ${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.0/include
```

but it is less explicit and I don't know if it will work.

Best regards,

Andrii

source/util.cpp
andriish: Hi @lyskov ,

I was surprised too.  But that was  on newest fedora (i.e. newest gcc 10.2.1  and LLVM  11.0.0), so maybe they are stricter.

It was here:
https://github.com/andriish/binder/runs/1026649950   <- Fedora33. Are you able to see these runs?

```
/github/workspace/source/util.cpp: In function ‘std::string binder::get_text(const clang::comments::Comment*, const clang::SourceManager&, clang::SourceLocation)’:
/github/workspace/source/util.cpp:258:71: error: could not convert ‘tc->clang::comments::TextComment::getText()’ from ‘llvm::StringRef’ to ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’}
  258 |  if( auto tc = dyn_cast<comments::TextComment>(C) ) return tc->getText();
      |                                                            ~~~~~~~~~~~^~
      |                                                                       |
      |                                                                       llvm::StringRef
```


Best regards,

Andrii

lyskov: got it, - thank you!

CMakeLists.txt
lyskov: aha, - i see your point. Lets use current solution then.

Issue #135, opened 2020-08-28
Hopefully last issue in CI


andriish:
Hi @lyskov ,

right now the Ci in master finds the following differences  between the generated files 

```
1_Binder_on_Fedora26.txt
2020-08-27T23:27:25.5181510Z 23/23-th strings are not equal /github/workspace/test/T01_enum.cpp /github/workspace/test/T01.enum.ref
2020-08-27T23:27:25.5182405Z    ->		.value("E1_V0", E1::E1_V0)<-
2020-08-27T23:27:25.5182659Z    ->		.value("E1_V0", E1_V0)<-
--
2020-08-27T23:27:25.5190711Z 51/51-th strings are not equal /github/workspace/test/T07_class.cpp /github/workspace/test/T07.class.ref
2020-08-27T23:27:25.5191139Z    ->			.value("E1V", Enums::E1::E1V)<-
2020-08-27T23:27:25.5191517Z    ->			.value("E1V", Enums::E1V)<-
1_Binder_on_Fedora27.txt
2020-08-27T23:30:47.7376241Z 23/23-th strings are not equal /github/workspace/test/T01_enum.cpp /github/workspace/test/T01.enum.ref
2020-08-27T23:30:47.7378065Z    ->		.value("E1_V0", E1::E1_V0)<-
2020-08-27T23:30:47.7378527Z    ->		.value("E1_V0", E1_V0)<-
--
2020-08-27T23:30:47.7400452Z 51/51-th strings are not equal /github/workspace/test/T07_class.cpp /github/workspace/test/T07.class.ref
2020-08-27T23:30:47.7401154Z    ->			.value("E1V", Enums::E1::E1V)<-
2020-08-27T23:30:47.7401569Z    ->			.value("E1V", Enums::E1V)<-
```
Only two systems are affected.  

It looks either ``getQualifiedNameAsString()`` works slightly different in different versions or 
``std::string bind_enum(std::string const & module, EnumDecl const *E)`` uses some assumptions that are not always true for different stl/llvm versions.

Best regards,

Andrii

andriish:
Fixed in master by #137

PR #136, opened 2020-08-28
Update t42

andriish:
Hi @lyskov ,

this is an update of reference files. 
There is also an update of diffbinder.cpp to take into account that LLVM11 had changed the names  of the types which makes binder produce slightly different codes (whitespace inserted), e.g. 
```
"C++: std::multiset<float, std::less<float>, std::allocator<float>>::key_comp() const --> struct std::less<float>");
```
vs.
```
"C++: std::multiset<float, std::less<float>, std::allocator<float> >::key_comp() const --> struct std::less<float>")
```

This MR makes the code comparison in CI work for Fedora28,29,30,31,32,33.

Best regards,

Andrii

P.S. If https://github.com/RosettaCommons/binder/issues/135 is solved, the comparison of codes in CI will actually work for all the systems.

### DIFF
--- a/test/T42.stl.names.multi.ref
+++ b/test/T42.stl.names.multi.ref
@@ -22,26 +22,34 @@
 
 void bind_std_stl_function(std::function< pybind11::module &(std::string const &namespace_) > &M)
 {
-	{ // std::binary_function file:bits/stl_function.h line:118
+	{ // std::binary_function file:bits/stl_function.h line:114
 		pybind11::class_<std::binary_function<float,float,bool>, std::shared_ptr<std::binary_function<float,float,bool>>> cl(M("std"), "binary_function_float_float_bool_t", "");
 		cl.def( pybind11::init( [](std::binary_function<float,float,bool> const &o){ return new std::binary_function<float,float,bool>(o); } ) );
 		cl.def( pybind11::init( [](){ return new std::binary_function<float,float,bool>(); } ) );
+		cl.def("assign", (struct std::binary_function<float, float, bool> & (std::binary_function<float,float,bool>::*)(const struct std::binary_function<float, float, bool> &)) &std::binary_function<float, float, bool>::operator=, "C++: std::binary_function<float, float, bool>::operator=(const struct std::binary_function<float, float, bool> &) --> struct std::binary_function<float, float, bool> &", pybind11::return_value_policy::automatic, pybind11::arg(""));
 	}
-	{ // std::equal_to file:bits/stl_function.h line:351
+	{ // std::equal_to file:bits/stl_function.h line:204
 		pybind11::class_<std::equal_to<float>, std::shared_ptr<std::equal_to<float>>, std::binary_function<float,float,bool>> cl(M("std"), "equal_to_float_t", "");
 		cl.def( pybind11::init( [](){ return new std::equal_to<float>(); } ) );
 		cl.def("__call__", (bool (std::equal_to<float>::*)(const float &, const float &) const) &std::equal_to<float>::operator(), "C++: std::equal_to<float>::operator()(const float &, const float &) const --> bool", pybind11::arg("__x"), pybind11::arg("__y"));
+		cl.def("assign", (struct std::binary_function<float, float, bool> & (std::binary_function<float,float,bool>::*)(const struct std::binary_function<float, float, bool> &)) &std::binary_function<float, float, bool>::operator=, "C++: std::binary_function<float, float, bool>::operator=(const struct std::binary_function<float, float, bool> &) --> struct std::binary_function<float, float, bool> &", pybind11::return_value_policy::automatic, pybind11::arg(""));
 	}
 }
 
 
 // File: std/functional_hash.cpp
-#include <bits/std_function.h> // std::_Manager_operation
-#include <bits/std_function.h> // std::bad_function_call
+#include <functional> // std::_Manager_operation
+#include <functional> // std::bad_function_call
 #include <functional> // std::hash
+#include <functional> // std::less
 #include <iterator> // __gnu_cxx::__normal_iterator
+#include <map> // std::_Rb_tree_color
+#include <map> // std::_Rb_tree_const_iterator
+#include <map> // std::_Rb_tree_iterator
+#include <map> // std::multimap
 #include <memory> // std::allocator
 #include <memory> // std::allocator_arg_t
+#include <set> // std::multiset
 #include <sstream> // __str__
 #include <stdexcept> // std::domain_error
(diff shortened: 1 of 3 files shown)


test/T42.stl.names.multi.ref
```diff
@@ -22,34 +22,26 @@
...
-		cl.def("assign", (struct std::binary_function<float, float, bool> & (std::binary_function<float,float,bool>::*)(const struct std::binary_function<float, float, bool> &)) &std::binary_function<float, float, bool>::operator=, "C++: std::binary_function<float, float, bool>::operator=(const struct std::binary_function<float, float, bool> &) --> struct std::binary_function<float, float, bool> &", pybind11::return_value_policy::automatic, pybind11::arg(""));
 	}
 }
 
 
 // File: std/functional_hash.cpp
-#include <functional> // std::_Manager_operation
-#include <functional> // std::bad_function_call
+#include <bits/std_function.h> // std::_Manager_operation
+#include <bits/std_function.h> // std::bad_function_call
```
lyskov, line 40: This looks wrong, could you please tell me what OS/LLVM version have you used to produce this reference results? I want to run this locally because it looks like include-resolution code will need adjustments.

test/T42.stl.names.multi.ref
```diff
@@ -22,34 +22,26 @@
...
 	}
-	{ // std::equal_to file:bits/stl_function.h line:204
+	{ // std::equal_to file:bits/stl_function.h line:351
 		pybind11::class_<std::equal_to<float>, std::shared_ptr<std::equal_to<float>>, std::binary_function<float,float,bool>> cl(M("std"), "equal_to_float_t", "");
 		cl.def( pybind11::init( [](){ return new std::equal_to<float>(); } ) );
 		cl.def("__call__", (bool (std::equal_to<float>::*)(const float &, const float &) const) &std::equal_to<float>::operator(), "C++: std::equal_to<float>::operator()(const float &, const float &) const --> bool", pybind11::arg("__x"), pybind11::arg("__y"));
-		cl.def("assign", (struct std::binary_function<float, float, bool> & (std::binary_function<float,float,bool>::*)(const struct std::binary_function<float, float, bool> &)) &std::binary_function<float, float, bool>::operator=, "C++: std::binary_function<float, float, bool>::operator=(const struct std::binary_function<float, float, bool> &) --> struct std::binary_function<float, float, bool> &", pybind11::return_value_policy::automatic, pybind11::arg(""));
```
lyskov, line 35: I think it would be good if we decide to always produce reference results on CentOS-7 with LLVM-6 because otherwise we will have meaningless stream of changes when  `assign` and constructor operators got deleted and inserted. 

Also, producing reference results on non-reference platform while not triggering `diffbinder` still do break `self-test.py` tests. So could you please regenerate results results on CentOS-7 with LLVM-6?

Thanks,

lyskov:
> P.S. If #135 is solved, the comparison of codes in CI will actually work for all the systems.
-- this is fantastic! I am looking forward to it @andriish !

test/T42.stl.names.multi.ref
andriish: Hi @lyskov ,
First,
> otherwise we will have meaningless stream of changes
>Also, producing reference results on non-reference platform ...
completely agree.


> regenerate results results on CentOS-7 with LLVM-6?
Done. And there is still a difference with the references.
Also, one can see that the current reference in master is problematic just from this include:

```
#include <bits/std_function.h> // std::_Manager_operation
```
it really should be 
```
#include <functional> // std::_Manager_operation
```

The ``bits/std_function.h`` include does not even exist on CentOS7 ( gcc-4.8.5) with clang/llvm-6.0. I mean at all.
 It comes from the gcc in devtoolset-7+. Most likely that happens because when you generate these references using devtoolset and
binder picks up the stl from there. 
(+I'm not sure w/o searching for docs, but everything in the  ``bits/`` directory is implementation-dependent, I think. )

Would that make sense to add the path of stl headers to the python scripts explicitely?


Best regards,

Andrii

test/T42.stl.names.multi.ref
andriish: That was RHEL8 
gcc -8.3.1
llvm 8.0.1

lyskov: Thank you! I will try to replicate this locally.

test/T42.stl.names.multi.ref
lyskov: > Would that make sense to add the path of stl headers to the python scripts explicitely?

-- this would not help. The issue is as you pointed out that some STL versions/implementations do have multiple layers of includes so "final" include for type/function/class will not necessary will be one that specified by C++ standard. We are addressing this in `add_relevant_include_for_decl` function (see 
https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L98) by mapping standard types to known C++11 includes and by mapping known non-standard include path into standard one. This approach however require adjustment from time to time as new failure cases arise.

test/T42.stl.names.multi.ref
```diff
@@ -303,12 +303,12 @@ PYBIND11_MODULE(T42_stl_names_multi, root_module) {
 
 }
 
-// Source list file: TEST/T42_stl_names_multi.sources
+// Source list file: /home/andriish/Projects/binder/test//T42_stl_names_multi.sources
```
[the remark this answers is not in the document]
andriish: Hi @lyskov ,
Done.
I just manually copy the generated results.

Best regards,
Andrii

test/T42.stl.names.multi.ref
andriish: OK. I saw it, but did not put enough attention to understand why that map is there.
Would that make it possible to use another, more automatic approach?

Like if there is a tree of includes, some includes should be made special == the includes from standard.  And for every funny internal type the appropriate #include  would be just "navigate in the include tree till the first ancestor that is stl  include".
Of course, there should be a list of stl includes, but that is kind of simple as it is in standard.

Best regards,
Andrii

lyskov: Yes, i was thinking about this too, but so far it remains unclear how to extract include-chain info from LLVM AST. If you know of the examples of how to do it please let me know. Include mapping problem still remain as one of the unsolved issues so i am eager to revisit this when info of how to extract include chains became available. Thanks,

andriish: I don't know yet. The LLVM  does not look that nice if to try to actually do  something.

test/T42.stl.names.multi.ref
lyskov: ok, fix for this is now in master

PR #138, opened 2020-09-03
Added clang-format config (for consideration)

andriish:
Hi @lyskov ,

here is an attempt to have a clang-format config that would match the observed formatting rules in the code (e.g. tabs).

If you find it reasonable, and  the `clang-format  -i source/*p` command  is executed on the reference platform once: 

 - the need to check or adjust formatting in any merge request will disappear. 

 - practically the formatting check can be integrated into CI. Again, no need for a manual check.

Also, the format file is not my personal preferences (don't have any), but an attempt to match the existing rules. The idea of this MR is just to have any formatting config so the code formatting can be done  (semi-)automatically.

Bets regards,

Andrii

P.S. I could not deduce the prefered style for the breaks before braces from the code. Sometimes it looks like "Attach", i.e. no break before the brace, sometimes it is "Allman", i.e. always break. Therefore left a commented line.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,18 @@
+---
+BasedOnStyle:  LLVM
+AlignTrailingComments: false
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: true
+ColumnLimit:     0
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+BreakConstructorInitializersBeforeComma: false
+Standard:        Cpp11
+IndentWidth:     4
+TabWidth:        4
+UseTab:          Always
+BreakBeforeBraces: Attach
+#BreakBeforeBraces: Allman
+SpacesInParentheses: true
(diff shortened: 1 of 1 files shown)


lyskov (approved):
Thank you for adding this @andriish !

Re using clang-format in CI: - i am not sure about this. I tried this a few times and overall results was rather mixing.

andriish:
Hi @lyskov ,

yes, the `clang-format` is not my first choice as well ( the first choice is a standalone `astyle`).  It requires the whole clang machinery and results vary from version to version.
However,   because binder requires clang anyway it is very logical to use clang-format.

So well, yes,  if you find it ok and run `clang-format -i source/*p` on any reference system, I will add CI with that reference system that checks the formating. 

Best regards,
Andrii

lyskov:
@andriish sorry if my previous message was cryptic, what i meant is: - i am not sure integrating (any) format checking into CI worth the trouble.

PR #145, opened 2020-10-22
Fix tests pybind11 v2.6

andriish:
Hi @lyskov ,

here are fixes for CI

1)  To prevent some tests from running with old pybind11 (<2.6.0).
The pybind11 2.6.0 appeared just 2 days ago, so the fix literally prevents tests from running is the version is insufficient. 
In some weeks/days the latest Fedoras will be updated and all tests will be active again for it.
The fix will still make sense for older distros.

2) It seems the regenerated references differ from the previous version so the  difference tests are now set to always pass.

Best regards,

Andrii

Another, but more global question:   would you consider to tag a next version of binder?


lyskov (approved):
Thank you for adding this @andriish !

PR #146, opened 2020-10-24
Adding support for compilation with external LLVM on Mac

andriish:
Hi @lyskov,

This PR:  
    - Adds support for compilation with external LLVM on Mac.
    - Adds CI for Mac. ( Only llvm 11).
    - Adds a bit of docs on compilation with external LLVM on Mac.

Some tests are disabled on mac because I have an old one. These should work fine on  a newer version.    
 

Best regards,

Andrii

### DIFF
--- a/.github/workflows/osx/entrypoint.sh
+++ b/.github/workflows/osx/entrypoint.sh
@@ -1,0 +1,11 @@
+#!/bin/sh -l
+set -x
+brew install wget coreutils
+wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+tar -Jxo clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+export PATH=$PATH:clang+llvm-11.0.0-x86_64-apple-darwin/bin
+
+cmake CMakeLists.txt
+make
+otool -L source/binder
+ctest . --output-on-failure 
--- a/.github/workflows/osx/entrypoint.sh
+++ b/.github/workflows/osx/entrypoint.sh
@@ -4,9 +4,6 @@ brew install wget coreutils xz
 wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
 ls
 tar -xJf clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
-#xz -d clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
-#ls
-#tar -x clang+llvm-11.0.0-x86_64-apple-darwin.tar
 export PATH=$PATH:clang+llvm-11.0.0-x86_64-apple-darwin/bin
 
 cmake CMakeLists.txt
--- a/.github/workflows/osx/entrypoint.sh
+++ b/.github/workflows/osx/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/bin/sh -l
 set -x
-brew install wget coreutils xz
+brew install wget coreutils xz pybind11
 wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
 ls
 tar -xJf clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
--- a/.github/workflows/osx/entrypoint.sh
+++ b/.github/workflows/osx/entrypoint.sh
@@ -2,7 +2,9 @@
 set -x
 brew install wget coreutils xz
 wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+ls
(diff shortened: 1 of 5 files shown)


.github/workflows/osx/entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -l
+set -x
+brew install wget coreutils xz pybind11
+wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+ls
+tar -xJf clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+export PATH=$PATH:clang+llvm-11.0.0-x86_64-apple-darwin/bin
+
+cmake CMakeLists.txt
+make
```
lyskov, line 10: This command will build whole set of executable from LLVM sources, but we only need Binder plus headers so maybe we can only build ` binder tools/clang/lib/Headers/clang-headers` targets? (this will be ~10x faster)

documentation/install.rst
```diff
@@ -138,6 +138,24 @@ For Ubuntu18+ run, an example for LLVM/Clang 10:
     sudo apt-get -y install  clang-10 llvm-10 libclang-10-dev llvm-10-dev
     sudo apt-get -y install  cmake make gcc g++
 
+For MacOSX:
+
+  Make sure the XCode is installed. If needed, install cmake, python and other utilities, e.g. using homebrew:
+  
+  .. code-block:: console
+    
+    brew install wget coreutils xz pybind11 cmake
```
lyskov, line 147: Upstream Pybind11 should not be used, instead please point user to use https://github.com/RosettaCommons/pybind11 . It have master that is in lock-step with Binder sources so it should always work correctly.

documentation/install.rst
```diff
@@ -138,6 +138,24 @@ For Ubuntu18+ run, an example for LLVM/Clang 10:
...
+
+
+  Download and install the llvm+clang from the official site, e.g. using ``wget`` and 
+  add the location of llvm config to the $PATH:
+
+  .. code-block:: console
+    
+    wget --no-verbose https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+    tar -xJf clang+llvm-11.0.0-x86_64-apple-darwin.tar.xz
+    export PATH=$PATH:$(pwd)/clang+llvm-11.0.0-x86_64-apple-darwin/bin
```
lyskov, line 157: Why do we need this special instructions? We already have this script: https://github.com/RosettaCommons/binder/blob/master/build.py which take care of installing Pybind11, LLVM and compiling the later...

test/CMakeLists.txt
```diff
@@ -101,6 +111,17 @@ if (pybind11_VERSION VERSION_LESS 2.5.99)
...
+ list(REMOVE_ITEM binder_tests T42.stl.names.multimap)
+ list(REMOVE_ITEM binder_tests T42.stl.names.multiset)
+ list(REMOVE_ITEM binder_tests T42.stl.names.set)
+ list(REMOVE_ITEM binder_tests T42.stl.names.map)
+ list(REMOVE_ITEM binder_tests T42.stl.names)
+ list(REMOVE_ITEM binder_tests T40.stl)
+ list(REMOVE_ITEM binder_tests T11.override)
+ list(REMOVE_ITEM binder_tests T10.inheritance)
+ list(REMOVE_ITEM binder_tests T02.function)
+endif()
```
lyskov, line 124: out of curiosity: - what exactly wrong with these tests on Mac? Is it something that can be fixed?

.github/workflows/osx/entrypoint.sh
andriish: Hi @lyskov ,

no, that uses the CMakeLists.txt from the top directory. It will build only the binder.

The CI works in the same way as on Linux.

documentation/install.rst
andriish: This is a precompiled LLVM.   
Please note that it would be too "expensive" to compile the whole LLVM for each commit.
With the precompilled LLVM the whole pipeline takes ~ 5 min.

test/CMakeLists.txt
andriish: There is a clash of different system headers. Something like pre C++11 vs post C++11. One should set the paths correctly and some flags to make is disapear.  But this problem  is relevant only for older macs, e.g. my working 10.13, things should work fine on 10.15.  I'm going to request a newer machine, but that is my tertiary system, so, well, I survive as it is.
 In short, that is not binder problem.

documentation/install.rst
lyskov: aha, got it i see bin at the end of PATH now :)

lyskov (approved):
LGTM! Thank you for adding this @andriish !

andriish:
Hi @lyskov,

thanks for merging!
What about tagging a next version?

Andrii

lyskov:
> What about tagging a next version?

-- what do you mean, could you please elaborate?

andriish:
Hi @lyskov ,

I mean https://github.com/RosettaCommons/binder/tags   
Having binder 2.0.0 or 1.1.0 or 1.0.1 etc.  
"New release".

Best regards,

Andrii

lyskov:
Good idea, it is more then year from the last release. Done, i created a new one: v1.1.0

andriish:
Thank you very much!

PR #147, opened 2020-12-07
Expanded support to anonymous union.

AzaelCicero:
Added support for anonymous union.


lyskov (approved):
Looks good to me, - thank you for putting this together @AzaelCicero !

PR #154, opened 2021-04-24
Fix reference codes

### LINKED ISSUE #153: test cases outdated
Hi @lyskov,
first of all: thanks a lot for this tool!

I just realised that many of the test case references seem to be out of date.
CTest does not report any discrepancies, but only because `--always-success` is being passed to `diffbinder`.  
This flag has been introduced in 902f6e7bec36ac4a5bae810d150c470626ea4b5e.
I think the flag needs to be removed and test cases updated. Do you agree?

P.S.: For most of the test cases, the differences can be resolved easily by removing some semi-colons - they are related to #152.
For some other test cases, however, like T42, the changes are more extensive.

andriish:
In this PR:

- Removed excessive semi-colon from the reference sources. Should close #153, tag @ThFe
- Added Fedora34 to the CI
- Output test results for Darwin

### DIFF
--- a/.github/workflows/fedora34/Dockerfile
+++ b/.github/workflows/fedora34/Dockerfile
@@ -1,0 +1,8 @@
+# Container image that runs your code
+FROM fedora:34
+
+#Copies your code file from your action repository to the filesystem path `/` of the container
+COPY entrypoint.sh /entrypoint.sh
+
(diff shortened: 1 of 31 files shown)


lyskov (approved):
LGTM, - thank you for taking care of this @andriish !

PR #157, opened 2021-05-09
Forward declared inner functions

### LINKED ISSUE #156: Forward declared inner functions
Hi,
I ran into a problem when binding classes with forward declared inner classes.
In particular, when binding the code
```
struct outer {
	struct inner { int a; };  // ok
	struct inner2;             // ok
	struct inner3;             // not ok

	void f( inner *i, inner2 *i2, inner3 *i3 ){};

	struct inner2 { int b; };
};

struct outer::inner3 { int c; };
```
the inner class `inner3` will not be included.
However, I would have expected bindings for `inner3` as well.
Do you agree?

ThFe:
This pull request should close #156.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -1080,6 +1080,11 @@ string ClassBinder::bind_nested_classes(Context &context)
 {
 	string c;
 	for_public_nested_classes([&c, &context, this](CXXRecordDecl const *innerC) {
+		if ( !innerC->isFirstDecl() ) return;
+		if ( !innerC->isThisDeclarationADefinition() && innerC->hasDefinition() ) {
+			// innerC seems to be a forward declaration
+			innerC = innerC->getDefinition();
+		}
 		if(is_bindable(innerC)) {
 			//c += "\t// Binding " + C->getNameAsString() + ";\n";
 			ClassBinder b(innerC);
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -1080,7 +1080,7 @@ string ClassBinder::bind_nested_classes(Context &context)
 {
 	string c;
 	for_public_nested_classes([&c, &context, this](CXXRecordDecl const *innerC) {
-		if ( !innerC->isFirstDecl() ) return;
+		if ( !innerC->isFirstDecl() ) return; // avoid multiple bindings: try to bind innerC only at first declaration
 		if ( !innerC->isThisDeclarationADefinition() && innerC->hasDefinition() ) {
 			// innerC seems to be a forward declaration
 			innerC = innerC->getDefinition();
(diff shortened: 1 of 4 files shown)


source/class.cpp
```diff
@@ -1080,6 +1080,11 @@ string ClassBinder::bind_nested_classes(Context &context)
 {
 	string c;
 	for_public_nested_classes([&c, &context, this](CXXRecordDecl const *innerC) {
+		if ( !innerC->isFirstDecl() ) return;
```
lyskov, line 1083: Could you please add comment for this line stating that it is need to skip forward declarations to avoid repeated bindings? Thanks,

test/CMakeLists.txt
```diff
@@ -21,14 +21,14 @@ else()
...
+      --root-module ${stestnamenodot} --prefix ${CMAKE_CURRENT_BINARY_DIR}/ ${${stestname}_config_flag} --single-file  --annotate-includes  ${CMAKE_CURRENT_BINARY_DIR}/${stestname}.hpp.include  
       -- -x c++ -std=c++11   -I . -I ${CMAKE_CURRENT_SOURCE_DIR}  -isystem /  -I ${CLANG_INCLUDE_DIRS}   -iwithsysroot${LibClang_INCLUDE_DIR}
       WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}" VERBATIM)
   else()
     ADD_CUSTOM_TARGET(target${stestnamenodot}cpp
     #The first two expressions below are for older clang
     COMMAND C_INCLUDE_PATH=${LibClang_INCLUDE_DIR}   CPLUS_INCLUDE_PATH=${LibClang_INCLUDE_DIR}  ${CMAKE_BINARY_DIR}/source/binder --bind ""  -max-file-size=100000
-      --root-module ${stestnamenodot} --prefix ${CMAKE_CURRENT_BINARY_DIR}/ ${${stestname}_config_flag} --single-file  --annotate-includes  ${stestname}.hpp.include  
+      --root-module ${stestnamenodot} --prefix ${CMAKE_CURRENT_BINARY_DIR}/ ${${stestname}_config_flag} --single-file  --annotate-includes  ${CMAKE_CURRENT_BINARY_DIR}/${stestname}.hpp.include
```
lyskov, line 31: This change set does not seems related to PR and i am guessing was committed by mistake, - could you please revert this?

lyskov (changes requested):
Thank you for taking care of this @ThFe ! Please see a few minor comments that will need to be addressed before this is merged. Thank you,

source/class.cpp
ThFe: Done!

test/CMakeLists.txt
lyskov: Thank you! 

re need to compile: this is unexpected, tagging @andriish who might have thoughts about this.

lyskov (approved):
LGTM, - thank you for taking care of this @ThFe !

PR #158, opened 2021-05-24
Inheritance of inner classes

ThFe:
Hi @lyskov,

Trying bind the code
```
struct outer {
	struct inner {
		int a;
	};
};

struct outer2 : public outer::inner {
	double b;
};
```
currently fails, giving the error `"ERROR: Could not find binder for type: "`.
The reason is that - while `outer::inner` is registered as a dependency of `outer2` - there is no `Binder` object for `outer::inner`. 
(Bindings for `outer::inner` are generated when processing struct `outer`, though.)

This pull request solves this problem by:

1. introducing an simple `Binder` class called `Dependency` which for a given type (like `outer::inner`) depends on the type (like `outer`) whose `Binder` object actually provides the bindings and
2. registers for inner classes appropriate `Dependency` objects with the `context` object.

Regards,
Thomas

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -12,8 +12,10 @@
 
 #include <binder.hpp>
 #include <class.hpp>
+#include <dependency.hpp>
 #include <function.hpp>
 #include <enum.hpp>
+#include <memory>
 #include <type.hpp>
 #include <util.hpp>
 
@@ -1091,6 +1093,10 @@ string ClassBinder::bind_nested_classes(Context &context)
 			b.bind(context);
 			c += b.code();  c += '\n';
 			prefix_code_ += b.prefix_code();
+			// make context aware of innerC
+			// (this registers innerC with context with C as a dependency)
+			BinderOP bop = std::shared_ptr<Binder>( new Dependency( innerC, C ) );
+			context.add( bop );
 		}
 	});
 
--- a/source/context.cpp
+++ b/source/context.cpp
@@ -225,7 +225,11 @@ void Context::bind(Config const &config)
 
 		outs() << "Generate bindings, pass " << pass << "...\n";
 
-		for(auto & sp : binders) {
+		// Context::add might invalidate iterators over binders
+		// when called inside b.bind(*this)
+		// => need to loop over a copy here
+		auto binders_ = binders;
+		for(auto & sp : binders_) {
 			Binder & b( *sp );
 			if( !b.is_binded()  and  b.bindable() and  b.binding_requested() ) {
 				//outs() << "Binding: " << b.id() /*named_decl()->getQualifiedNameAsString()*/ << "\n";
--- a/source/dependency.hpp
+++ b/source/dependency.hpp
@@ -1,0 +1,56 @@
+// -*- mode:c++;tab-width:2;indent-tabs-mode:t;show-trailing-whitespace:t;rm-trailing-spaces:t -*-
+// vi: set ts=2 noet:
+//
+// All rights reserved. Use of this source code is governed by a
+// MIT license that can be found in the LICENSE file.
+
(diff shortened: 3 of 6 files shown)


source/class.cpp
```diff
@@ -1091,6 +1093,10 @@ string ClassBinder::bind_nested_classes(Context &context)
 			b.bind(context);
 			c += b.code();  c += '\n';
 			prefix_code_ += b.prefix_code();
+			// make context aware of innerC
+			// (this registers innerC with context with C as a dependency)
+			BinderOP bop = std::shared_ptr<Binder>( new Dependency( innerC, C ) );
+			context.add( bop );
```
lyskov, line 1099: This is problematic: right now Binder design is assume that there is a clear separation of parsing phase and generation phase. So allowing generation phase to add new Binders will break this. 

Let me think if there is a better way to handle this situation which preserve clear separation of parsing and generation phases

source/context.cpp
```diff
@@ -225,7 +225,11 @@ void Context::bind(Config const &config)
 
 		outs() << "Generate bindings, pass " << pass << "...\n";
 
-		for(auto & sp : binders) {
+		// Context::add might invalidate iterators over binders
+		// when called inside b.bind(*this)
+		// => need to loop over a copy here
+		auto binders_ = binders;
```
lyskov, line 231: This is problematic from performance perspective: for instance in my test project `binders` will hold ~50k objects so copying it will take notable time. So i would recommend to instead refactor this loop to use plain indexes.

lyskov (commented):
Thank you for putting this together @ThFe ! I left a few technical comments below, but before we jump to them could you please outline why allowing inheritance from an inner classes is 
sounds design? 

The way i see it, inner classes describe types which in way auxiliary for main class. This is of course subjective and one can argue that since C++ allow such inheritance then we should support it. So i am curious: could you please provide class design example that require inheritance from an inner class and which is a _good_ design? Thanks,

ThFe:
Hi @lyskov,

[DiscountingSwapEngine](https://github.com/lballabio/QuantLib/blob/master/ql/pricingengines/swap/discountingswapengine.hpp) gives an example of such a design (inheriting from an inner class).
Whether this is an example of a _good design_ is another question...

In general, I do share your concerns regarding such a design. On the other hand, C++ supports it and binding code for inner classes is already generated by binder (e.g., [T15.inner_class.hpp](https://github.com/RosettaCommons/binder/blob/master/test/T15.inner_class.hpp)). I therefore thought it good to have binder supporting this as well (I am currently facing this problem in a project where I do not have influence on the design).

In the meantime, I realised that the proposed solution may lead to Python import errors when two or more classes have identically named inner classes (the Python name would be the same then, leading to import errors).

In order to resolve this problem, I would propose to change binder such as to

1. not binding inner classes by default (needs to be required via `+class` config settings) and
2. allowing custom python class names (e.g., by adding an optional argument to  `+class ClassName [PythonName]`).

First point would change binder's behaviour, second point could be useful as well for templated classes.

[I could provide a proposal for the required code changes.]

source/class.cpp
ThFe: Ok. That was not clear to me (since context does provide the `add` method.)

lyskov:
@ThFe thank you for info! I agree that it would be preferable to be able generate bindings for everything that C++ technically supports.

re 1, 2: this might work for small/medium projects but for large code bases it might not be practical to require explicit listing of classes. Hm... it looks like the core issue here is that right now inner classes is bound when umbrella class is bound. So making inner classes bindable independently would solve this nicely. Let me think about this for a bit, maybe there is an elegant way to do that.

lyskov:
Not clear how to effectively implement solution for provided example, closing for now.

PR #168, opened 2021-12-06
Added missing include

andriish:
Added missing include.

### DIFF
--- a/test/T42.stl.names.map.hpp
+++ b/test/T42.stl.names.map.hpp
@@ -20,6 +20,7 @@
 #include <map>
 #include <unordered_map>
 
+#include <string>
 #include <list>
 #include <vector>
 #include <deque>


lyskov:
Awesome, - thank you for taking care of this @andriish ! Interestingly this has not triggered on rest of systems, i guess due to different include tree's in `std::` lib.

Regarding this particular failure i am curious: - have exactly have you found this? Was this somewhere in CI logs that i missed? Thanks,

lyskov:
oh never mind: i see that i missed this in logs 🤦🏻‍♂️:
https://github.com/RosettaCommons/binder/runs/4317223685?check_suite_focus=true#step:3:1096
```./T42.stl.names.map.hpp:33:24: error: no member named 'string' in namespace 'std'
void foo(std::map<std::string, std::list< std::forward_list< std::deque< std::vector<double> > > > >) {}
  ```

andriish:
Hi @lyskov ,

on std::string -- yes, the versions are slightly different and the newer  libraries are stricter.

release: awesome! Thanks!

Best regards,

Andrii

PR #171, opened 2021-12-26
Add handling for bits/stl_uninitialized.h

andriish:
Add handling for bits/stl_uninitialized.h, which appeared in some generated sources.
According to gcc docs this header should be related to <memory>.


lyskov (approved):
LGTM, - thank you @andriish !

PR #176, opened 2022-03-10
Doc: between 'make' and 'make install', run 'ctest'

jwuttke:
To help people detecting errors early on.


lyskov:
Looks good to me (assuming that CI tests pass), - @andriish what do you think about this change?

lyskov (approved):
wait, i just realized that proposed changes is to documentation not to cmake files, - sure this sounds good to me, - thank you!

PR #180, opened 2022-03-13
In terminal output, print input files we are processing

jwuttke:
In particular, printing the .hpp input files that are to be processed is necessary to explain repetative output like
```
Generate bindings, pass 1...
```

### DIFF
--- a/source/context.cpp
+++ b/source/context.cpp
@@ -228,7 +228,7 @@ void Context::bind(Config const &config)
 		for(auto & sp : binders) {
 			Binder & b( *sp );
 			if( !b.is_binded()  and  b.bindable() and  b.binding_requested() ) {
-				//outs() << "Binding: " << b.id() /*named_decl()->getQualifiedNameAsString()*/ << "\n";
+                            outs() << "Binding: " << b.id() /*named_decl()->getQualifiedNameAsString()*/ << "\n";
 				b.bind(*this);
 				flag=true;
 			}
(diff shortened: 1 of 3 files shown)


source/context.cpp
```diff
@@ -228,7 +228,7 @@ void Context::bind(Config const &config)
 		for(auto & sp : binders) {
 			Binder & b( *sp );
 			if( !b.is_binded()  and  b.bindable() and  b.binding_requested() ) {
-				//outs() << "Binding: " << b.id() /*named_decl()->getQualifiedNameAsString()*/ << "\n";
+                            if( O_verbose ) outs() << "Binding: " << b.id() /*named_decl()->getQualifiedNameAsString()*/ << "\n";
```
lyskov, line 231: This could lead to _a lot_ of output on any non-trivial project. I do agree however that this could be useful. 🤔 How about controlling this with different command-line-option? Not sure what the good name will be for it though, maybe `--verbose-print-declarations`?

lyskov (commented):
Looks good to me, - thank you for putting this together @jwuttke ! Please see minor comment below. Thanks,

source/context.cpp
jwuttke: In my understanding, "verbose" means just this: lots of output. This is why a command-line switch is needed. So I'd propose to resolve this in the opposite way, by demoting some other uncritical one-line outputs from verbose to regular, thereby reserving the "-v" flag to cases where it really makes a difference for the user whether or not the terminal is flooded with output.

If however you opt for three rather than two verbosity levels, then why not take inspiration from a number of other software projects [1] that have quiet (-q), normal, verbose (-v).

[1] https://unix.stackexchange.com/questions/191254

lyskov: I just checked amount of output and it actually not that bad since only name of the classes is printed (not all member functions etc). For fairly large project (~3M LOC) i am getting only ~14k lines which is quite reasonable. So let's keep things simple and merge this as-is.

PR #184, opened 2022-03-13
source: correct typos in function comments;

jwuttke:
try to make sense of unclear comments;

uniformly use '///' for Doxygen function/class/namespace comments


lyskov (approved):
Awesome, - thank you for for fixing this @jwuttke ! Looks good to me!

PR #196, opened 2022-04-12
Dockerize

zwimer:
Building docker can sometimes be tough and current docs only use the old Clang 6. Clang has since changed a lot including where releases are posted, names of tarballs, etc. The current `build.py` script will not work for newer Clang, and figuring out how to build it can be a bit of a challenge as a result.

This PR does 2 things:
1. It adds a `Dockerfile` which supports newer clang versions; binder can be run dockerized, one can simply mount a config and the directory to edit to do so.
2. The `Dockerfile` provides some level of documentation for those who wish to build binder; as it is essentially a step by step instruction list.


documentation/install.rst
```diff
@@ -222,19 +222,25 @@ If ``binder`` was build with some older versions of LLVM, one could also set the
 
 .. _building-static:
 
+With Docker
+***********
+
+A `Dockerfile` for building binder can be found in the ``binder`` repository linked here: https://github.com/RosettaCommons/binder/
```
lyskov, line 228: Probably best to use relative path here since someone could be using local-clone or fork of this repository. Could you please update this? Thanks,

lyskov (changes requested):
Looks good to me, - thank you for adding this @zwimer ! One small thing, could you please move Docker file into `examples` dir? Thanks,

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Build dependencies
+RUN apt-get update
+RUN apt-get install -yq \
+	"libclang-${CLANG_VERSION}-dev" \
+	cmake \
+	git
+
+# Clone binder source
+ARG REPO="https://github.com/RosettaCommons/binder.git"
+RUN git clone --depth 1 --branch "${BRANCH}" "${REPO}" /binder
```
lyskov, line 35: Unless i am mistaken using `--depth 1` will lead to issue if user tries to use non-head commits. Also, whole Binder repository is minuscule compare to LLVM clone size. So this option create more of inconvenience than a real space saver, so perhaps this option should removed?

zwimer: There is no following git checkout; since the exact commit being cloned is what is built I have `--depth 1` as it just saves time. I'm not dead set on it, but there should be only disadvantages to removing it unless we also add lines like:
```
ARG COMMIT=HEAD
RUN git checkout "${COMMIT}"
```
Would you prefer those added?

documentation/install.rst
zwimer: I'm not exactly familiar with whatever `.rst` is. I presume you mean replace `https://github.com/RosettaCommons/binder/` with `SOMETHING` ? If you tell me what the relative path would be I can replace it in the PR though, though I'm not sure what it relative to exactly since this will be hosted on readthedocs, no? Should the replacement string be `..` or something else?

Dockerfile
lyskov: I guess my initial comment was a bit cryptic, let me clarify: what i was referring to was the case when someone going to work on Binder development using this image as base for environment. Or for example if someone get into situation when s/he want to try different Binder builds. 

Please note that this comment was optional since we just provide an example of Dockerfile so i am happy to merge this with `depth` option set.

documentation/install.rst
lyskov: oh, i was not suggesting to make this into real link (it will be challenging since documentation could be on different web site). Ok, how about just replacing this link with `https://github.com/RosettaCommons/binder/examples` (assuming that we are moving Dockerfile into examples dir). Thanks,

Dockerfile
zwimer: I see. Since I'll be moving this into the examples dir I suspect it would be better without that. If you do desire to make a `dev` example, I think you should instead do a `COPY ./binder /binder` so that the devs would not need to pull from a remote source every build (which also means testing small commits would not need to have a git push each time).

lyskov: Hm... i guess `COPY` would make more sense but in the same time using `git` will allow one to use Dockerfile on its own. I think i prefer later a bit more though. Let's accept that things is not 'fully-optimal' here and just merge this, - after all this suppose to. only serve as example on which one could build on.

lyskov (approved):
LGTM, - thank you for adding this @zwimer !

Issue #192, opened 2022-04-02
std/complex.cpp does not compile


jwuttke:
Same test code as in issue #191, corrected so that all auto-generated code is to be compiled. Compilation fails with errors from deep within pybind11. Error messages under clang (under gcc same errors, but even more akward messages):
```
[ 75%] Building CXX object CMakeFiles/BoundSand.dir/auto/std/complex.cpp.o
In file included from /G/int/binder-sandbox/complex/auto/std/complex.cpp:9:
In file included from /usr/include/pybind11/pybind11.h:13:
In file included from /usr/include/pybind11/detail/../attr.h:13:
/usr/include/pybind11/cast.h:653:5: error: static_assert failed due to requirement 'std::is_base_of<pybind11::detail::type_caster_base<std::complex<float>>, pybind11::detail::type_caster<std::complex<float>, void>>::value' "Holder classes are only supported for custom types"
    static_assert(std::is_base_of<base, type_caster<type>>::value,
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/pybind11/cast.h:722:48: note: in instantiation of template class 'pybind11::detail::copyable_holder_caster<std::complex<float>, std::shared_ptr<std::complex<float>>>' requested here
class type_caster<std::shared_ptr<T>> : public copyable_holder_caster<T, std::shared_ptr<T>> { };
                                               ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/type_traits:1422:66: note: in instantiation of template class 'pybind11::detail::type_caster<std::shared_ptr<std::complex<float>>>' requested here
    : public integral_constant<bool, __is_base_of(_Base, _Derived)>
                                                                 ^
/usr/include/pybind11/cast.h:762:5: note: in instantiation of template class 'std::is_base_of<pybind11::detail::copyable_holder_caster<std::complex<float>, std::shared_ptr<std::complex<float>>>, pybind11::detail::type_caster<std::shared_ptr<std::complex<float>>>>' requested here
    std::is_base_of<detail::type_caster_holder<base, holder>, detail::type_caster<holder>> {};
    ^
/usr/include/pybind11/detail/common.h:747:49: note: in instantiation of template class 'pybind11::detail::is_holder_type<std::complex<float>, std::shared_ptr<std::complex<float>>>' requested here
[...]

PR #198, opened 2022-05-05
add std::complex to python builtins

jwuttke:
As discussed in issue #192.

Demo code reported in that issue does compile and works correctly.

Still hoping for native, fast implementation though.

### DIFF
--- a/source/type.cpp
+++ b/source/type.cpp
@@ -612,6 +612,7 @@ bool is_python_builtin(NamedDecl const *C)
 													  "std::__hash_value_type",
 
 													  "std::function",
+                                                                                                          "std::complex"
  	};
 
 	for(auto &k : known_builtin) {


lyskov:
Looks good to me, - thank you @jwuttke! 

> Still hoping for native, fast implementation though

-- could you please elaborate on this? Or do you mean you want to be able to generate bindings for `std::complex` with Binder instead of using Pybind11? If so: i am not sure if that will be noticeably faster. Sure we will be able to modify `complex` in-place, instead of copying it, but copy operation in this case will only be copying ~16 (or 32 in case of long double) bytes. So I am a skeptical that this will take considerable time (comparing to cost of doing any numeric processing in Python).

jwuttke:
Thank you, Sergey.

> > Still hoping for native, fast implementation though
> -- could you please elaborate on this? Or do you mean you want to be able to generate bindings for `std::complex` with Binder instead of using Pybind11?

Yes, this I meant, or so at least I understood alternative 1 in our discussion at https://github.com/RosettaCommons/binder/issues/192. If there is no significant speed penalty with the current solution, then all the better. So I closed #192.

PR #202, opened 2022-05-17
spell out namespace prefixes, rather than applying 'using' to entire namespaces.

jwuttke:
`using namespace clang|llvm` may be convenient in coding, but is painful for someone who wants to understand the inner working of Binder. For such newcomer, it is very helpful to point more explicitly to where classes and functions are defined.

### DIFF
--- a/source/binder.cpp
+++ b/source/binder.cpp
@@ -33,9 +33,15 @@
 #include <class.hpp>
 #include <util.hpp>
 
-using namespace clang::tooling;
-using namespace llvm;
+namespace ct = clang::tooling;
+namespace cl = llvm::cl;
 
+using llvm::outs;
+using llvm::errs;
+
+using std::string;
+
+using binder::Config;
 
 // Apply a custom category to all command-line options so that they are the
 // only ones displayed.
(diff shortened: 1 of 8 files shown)


lyskov (commented):
Hm... i am not sure i see the need to this refactoring at least not on proposed scale. Sure using `using` declarations in headers (which we don't) is problematic but using them in `.cpp` sounds quite reasonable to me, particularly when there is an easy way to locate types definition (more on this below). 

@jwuttke could you please elaborate on what exactly do you think the issue with current code? You have mentioned that there is a difficulty of finding where is declarations coming from? What i usually do is to simply google something like `llvm` + `<class name>` which instantly takes me to LLVM autogenerate docs (like lvm+RecursiveASTVisitor → 
 https://clang.llvm.org/doxygen/classclang_1_1RecursiveASTVisitor.html). Or just navigate LLVM doxygen pages directly. Have you tried this? Thanks,

jwuttke:
"what exactly do you think the issue with current code?" - I wouldn't say there are issues, but I do think the readability of the code could be much improved by a number of relatively small measures.

This proposal to spell out namespace members or namespace prefixes instead of "using namespace" is just one of them, and not the most important one in my eyes. For other proposals, please see https://github.com/RosettaCommons/binder/discussions/200. I am looking forward for your opinion on these.

Regarding this PR about namespaces: When reading code, about the first thing I want to know is where certain identifiers come from: are they locally defined, are they class members, are they project specific, or do they come from an external API. Of course I can google them, or look up the API docs of the included libraries. What an effort though. Explicit namespace prefixes give the answer by orders of magnitude faster.

lyskov:
I thought LLVM names kind of standout so it is clear what project name belong to. But this could be quite idiosyncratic so i do see your point. I will be happy to merge some (maybe most?) of these changes so we can make it more clear when names come from. (i am guessing this was closed due to #200 related PR's?)

jwuttke:
Thank you for giving this a second thought and a new chance. Let's first get the other pending PRs done, then I will rebase and reopen this one.

PR #203, opened 2022-05-20
ran clang-format over all *.hpp and .cpp files

jwuttke:


lyskov:
Thank you for taking care of this @jwuttke ! I am planning to take more throughout look in a bit.

Re files in `fmt` dir: this is a third party library so i think it would be best to keep it formatted as-is. Thanks,

jwuttke:
ok, redone without source/fmt

jwuttke:
How conservative are you about C++ standard? With C++20, we will finally get std::format.

lyskov:
@jwuttke while we at this: would you be open to tweak clang-format options a bit? (i noticed a few oddities that probably best to take care now then later). Please let me know if can and i will outline problematic formatting. Thanks,

jwuttke:
Absolutely. It's quite esay to rerun clang-format, and force push again. So choose your favorite clang-format settings, push .clang-format, and let me know.

My favorites would be `UseTab: Never` and `ColumnLimit: 100`, but I will not insist any further on these sensitive preferences.

jwuttke:
I just noticed that current .clang-format allows no more than one consecutive blank line. I guess you would prefer `MaxEmptyLinesToKeep: 2` - not to mention 3...

lyskov:
Great! Let me play with formatting options locally and i will get back to you.

Re tabs-vs-spaces i guess i should have mentioned this earlier: Binder was designed to generate Python bindings for Rosetta which is ~3M LOC and for historical reasons using tabs for indentation. And since Binder is part of Rosetta it will be best to keep tabs for consistency.

lyskov:
@jwuttke i have pushed new version of `.clang-format` file. Could you please pull from master and re-format code? Thanks,

jwuttke:
Which version of clang-format are you using? I get errors from 11.0 and from 13.0.

lyskov:
> Which version of clang-format are you using? I get errors from 11.0 and from 13.0.

I am using v14.0

jwuttke:
There is just one incompatibility with clang-13: `SpaceBeforeParens: Custom` is not yet supported. If you could give up that specific option, it would be much, much easier for me to contribute.

lyskov:
OK, i pushed the new version of `.clang-format` file which should work fine with v13. Could you please give it a try? Thanks,

jwuttke:
Thank you, and sorry: I had overlooked an error message from the line `PackConstructorInitializers: CurrentLine`. After outcommenting it, I am now able to run clang13 over all source files, including under test/ and examples/.

source/context.cpp
```diff
@@ -360,8 +360,8 @@ PYBIND11_MODULE({1}, root_module) {{
 
 
 const char *module_header = R"_(
-#include <pybind11/pybind11.h>
 #include <functional>
+#include <pybind11/pybind11.h>
```
lyskov, line 364: hm... it is odd that these include blocks got rearranged, - i guess different clang-format versions might have some discrepancies...

lyskov (approved):
LGTM, - thank you for taking care of this @jwuttke !

source/context.cpp
jwuttke: It's good to have them rearranged in standard form - even if clang13 is wrong about touching code in quotes. Assuming that this is corrected in clang14, I would expect that clang14 won't revert these changes. Can you confirm?

PR #204, opened 2022-05-20
New source files main.cpp, options.hpp; move IncludeSet and Binder from context.hpp to binder.hpp

jwuttke:
As proposed in https://github.com/RosettaCommons/binder/discussions/200.

This PR depends on the previous one, https://github.com/RosettaCommons/binder/pull/203; i.e. it is meant to be rebased on master after merger of #203.

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -26,11 +26,11 @@
 #include <llvm/Support/CommandLine.h> // Declares llvm::cl::extrahelp
 
 #include <class.hpp>
+#include <config.hpp>
 #include <context.hpp>
 #include <enum.hpp>
 #include <function.hpp>
 #include <options.hpp>
-#include <util.hpp>
 
 using namespace clang::tooling;
 using namespace llvm;
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -6,7 +6,7 @@
 // All rights reserved. Use of this source code is governed by a
 // MIT license that can be found in the LICENSE file.
 
-/// @file   binder/binder.cpp
+/// @file   binder/main.cpp
 /// @brief  Main
 /// @author Sergey Lyskov
 
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -23,8 +23,6 @@
 #include <clang/Basic/SourceLocation.h>
 #include <clang/Frontend/CompilerInstance.h>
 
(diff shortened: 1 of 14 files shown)


jwuttke:
All commits reenacted with new formatting. Now ready for review or merge.

lyskov:
@jwuttke just to double check: this PR _only splits code_, no other refactoring has been done, correct? Thanks,

jwuttke:
Yes, you have my word.

source/main.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+cl::opt<bool> O_verbose("v", cl::desc("Increase verbosity of output"), cl::init(false), cl::cat(BinderToolCategory));
+
+cl::opt<bool> O_suppress_errors("suppress-errors",
+								cl::desc("Suppres all the compilers errors. This option could be useful when you want to tell Binder to ignore non-critical errors (for example due to missing "
+										 "includes) and generate binding for part of code that Binder was able to parse"),
+								cl::init(false), cl::cat(BinderToolCategory));
+
+cl::opt<bool> O_flat("flat", cl::desc("When specified generated files into single directory. Generated files will be named as <root-module>.cpp, <root-module>_1.cpp, <root-module>_2.cpp, ... etc."),
+					 cl::init(false), cl::cat(BinderToolCategory));
```
lyskov, line 85: Since you created `options.hpp` then i think it will be very logical to move option definition into `option.cpp` for consistency (otherwise it is not clear to person reading header files where our options actually defined). Thanks,

lyskov (changes requested):
Looks good to me @jwuttke , - thank you for refactoring this! Please see minor comment below. Other then that i think it is ready to merge. Thanks,

source/main.cpp
jwuttke: Sounds plausible and easy for the five `O_...` variables declared in `options.hpp`. 

What about the other `O_...` variables that are set in `main.cpp`? Most of them seem to be used only locally, in the BinderVisitor constructor.

jwuttke: `BinderToolCategory` is used by the `main` function. So it has to be declared `extern` anyway.

Nonetheless, the solution with all `O_...` variables moved to `options` looks good to me. Will now push commit.

lyskov (approved):
Looks great to me, - thank you @jwuttke !

PR #205, opened 2022-06-01
Handle non-const copy constructors

chrkr:
I have added a little bit of code to handle an annoying case where a third-party library did not declare a copy constructor const. Binder generates a binding for a copy construtor with a const parameter regardless, and this necessitates manually fixing the generated bindings so have the bindings code compile properly.

This addition checks whether the constructor is actually const or not, and if it is not, the generated bindings also do not declare the parameter const. If it is const, the generated bindings are as they were before.


lyskov (approved):
LGTM!

PR #207, opened 2022-06-07
Invert logic of some conditionals to save indentations

jwuttke:


source/class.cpp
```diff
@@ -160,7 +160,7 @@ string bind_data_member(FieldDecl const *d, string const &class_qualified_name_)
 	}
 
 	if( d->getType().isConstQualified() ) return ".def_readonly(\"{}\", &{}::{})"_format(d->getNameAsString(), class_qualified_name, d->getNameAsString());
-	else return ".def_readwrite(\"{}\", &{}::{})"_format(d->getNameAsString(), class_qualified_name, d->getNameAsString());
+	return ".def_readwrite(\"{}\", &{}::{})"_format(d->getNameAsString(), class_qualified_name, d->getNameAsString());
```
lyskov, line 163: @jwuttke could you please argue why the proposed style is better compare to original layout? Thanks,

jwuttke: LLVM style guide: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return.

However, quick web search shows that there is also some opposition to this rule.

Up to you to decide. I can of course reset this PR to the second but last commit.

lyskov: Hm... i am not sure about this. Also, it looks like this PR is trying to use this guide line: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code which i also find depending on particular coding taste.

lyskov:
About other changes: - let me think about this but right now i am not sure that merging this will improve readability the way i see it. Particularly early return/continue coupled with testing of negated expressions (ie https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code) make things harder to read IMHO.

source/config.cpp
```diff
@@ -84,115 +84,107 @@ void Config::read(string const &file_name)
...
 
-			if( line.back() == '\r' ) {
-				line.pop_back();
-				if( line.empty() ) continue;
+		if( line.back() == '\r' ) {
+			line.pop_back();
+			if( line.empty() ) continue;
+		}
+
+		if( !(line[0] == '+' or line[0] == '-') ) throw std::runtime_error("Invalid token at the begining of line in config file! Each line should begin with ether '+' or '-' or '#'! Line: " + line);
```
lyskov, line 94: `if( line[0] != '+' and line[0] != '-') ...`

source/config.cpp
```diff
@@ -84,115 +84,107 @@ void Config::read(string const &file_name)
...
+		size_t space = line.find(' ');
+		if( space == string::npos )
+			throw std::runtime_error("Invalid line in config file! Each line must have token separated with space from object name. For example: '+function aaa::bb::my_function'. Line: " + line);
```
lyskov, line 97: since this does not fit on single line then we have to use brackets or it became error prone (say someone confuse this with Python and start add more lines after `if`, ie:

```
if( space == string::npos ) {
    throw std::runtime_error("Invalid line in config file! Each line must have token separated with space from object name. For example: '+function aaa::bb::my_function'. Line: " + line);
}
```

source/context.cpp
```diff
@@ -341,102 +340,101 @@ void Context::generate(Config const &config)
 
 	outs() << "Writing code...\n";
 	for( uint i = 0; i < binders.size(); ++i ) {
-		if( /*binders[i]->is_binded()  and*/ binders[i]->code().size() ) {
-
-			string np, file_name;
+		if( !binders[i]->code().size() ) continue;
```
lyskov, line 343: `if( binders[i]->code().empty() ) continue;`

lyskov (changes requested):
@jwuttke i like this change (looking in split-view mode made all the difference clear!) - thank you for taking care of this! Please see a few minor comments/suggestions below. Thanks,

source/config.cpp
jwuttke: ok, done

source/config.cpp
```diff
@@ -84,115 +84,107 @@ void Config::read(string const &file_name)
 	string line;
 
 	while( std::getline(f, line) ) {
-		if( line.size() ) {
-			if( line[0] == '#' ) continue;
+		if( !line.size() or line[0] == '#' ) continue;
```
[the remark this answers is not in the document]
jwuttke: ok, done

source/config.cpp
jwuttke: ok, done

source/context.cpp
jwuttke: ok, done

lyskov (approved):
LGTM, - thank you for taking care of this @jwuttke !

PR #210, opened 2022-06-30
Add basic Eigen support.

hogabrie:


hogabrie:
Sorry wrong target :see_no_evil:

Issue #214, opened 2022-08-16
Multiple questions associated to usage of binder with a large heavily templated project


kliegeois:
First of all, thanks for developing this tool; I have been able to successfully apply it in the past week on some projects.

I have couple of questions associated to its application on the project Trilinos ( https://github.com/trilinos/trilinos ), a large C++ project which relies heavily on some template parameters.

Quick context on project: 
- it is composed of several subprojects, called packages, which have some dependencies between them,
- each package is associated to a different team,
- all the packages rely heavily on template.

The installation of the library is as follows:
- in the source directory there are some .hpp and .cpp files, CMake, based on the options of the ETI, generates some .cpp files that are created in the build directory. Some of the files of the source directory are not copied to the build directory.
- the library is built using the CMake generated .cpp files and .hpp files and .cpp files written by the developers.
- the .hpp files and .so libraries are copied to the install directory.

The goal to use binder is to have an automatic way of creating and maintaining  a python interface for the project without requiring the developers of the different packages to maintain the interface themself.

I have the following questions:
1. In the config.cfg file it is possible disable one class or one function one by one; is it possible to disable every classes and functions and then enable some of them one at a time? If so, how can I do that?
2. For our templated functions/classes, we use some ETI strategies to build and link the library. Typically, we rely on CMake to generate some .cpp files in the build directory which are then compiled and linked to the .so lib. How can binder be aware of those ETI? How can we instantiate multiple times the same class with different template values and generate different Python variable names?
3. Currently, we use macros for the ETI such as:
```
#  define TPETRA_MULTIVECTOR_INSTANT(SCALAR,LO,GO,NODE) \
  template class MultiVector< SCALAR , LO , GO , NODE >; \
  template MultiVector< SCALAR , LO , GO , NODE > createCopy( const MultiVector< SCALAR , LO , GO , NODE >& src);
```
that we use as follows in a cpp file generated by CMake:
```
TPETRA_DETAILS_MAKECOLMAP_INSTANT( int, longlong, Kokkos_Compat_KokkosSerialWrapperNode )
```
Can binder work around the macros in this context and understand the fact that the .cpp files do the ETI?

4. In the source codes, we have some classes, functions, and includes that are guarded by some preprocessing checks.
For example, if some TPL is included some classes can be declared and defined, otherwise it is not the case.
When calling binder, is it possible to process those `#ifdef` checks and drop the content of the .hpp file between the begin and the end of the check if not defined?
[...]

PR #217, opened 2022-08-30
Add custom shared pointer option

kliegeois:
This PR is associated to one of the questions of #214.

The PR adds an option to specify a custom shared pointer class.
The header files can not be included into binder source code as it is unknown at compilation time.
Therefore, the proposed code relies on some string logic to avoid to have a custom shared pointer of a custom shared pointer.
The header(s) associated to the shared pointer should be specified with the `+include` option.
The proposed implementation avoids to create custom shared pointer of std classes.
Finally, the proposed implementation only supports to specify one custom shared pointer class.

Please @lyskov, let me know any comments/modifications that you would like to see in this PR.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -868,21 +868,14 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
 
 	string const include = relevant_include(C);
 
-	string shared_ptr = "std::shared_ptr";
-	bool use_custom_shared = Config::get().use_custom_shared();
-	if( use_custom_shared )
-		shared_ptr = Config::get().custom_shared();
-
+	string holder_type = Config::get().holder_type();
 
 	string c = "\t// Forward declaration for: " + qualified_name + " file:" + (include.size() ? include.substr(1, include.size() - 2) : "") + " line:" + line_number(C) + "\n";
 
-	string maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
+	string maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
 	//Check if the type is a custom shared pointer:
-	if( use_custom_shared && qualified_name.rfind(shared_ptr, 0) == 0 ) {
-		maybe_holder_type = "";
-	}
-	else if( is_inherited_from_enable_shared_from_this(C) ) {
-		maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
+	if( is_inherited_from_enable_shared_from_this(C) ) {
+		maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
 	}
 	else if( CXXDestructorDecl *d = C->getDestructor() ) {
 		if( d->getAccess() != AS_public ) maybe_holder_type = ", " + qualified_name + '*';
@@ -1184,19 +1177,12 @@ void ClassBinder::bind(Context &context)
 	string const trampoline_name = callback_structure_constructible ? callback_structure_name(C) : "";
 	string const binding_qualified_name = callback_structure_constructible ? callback_structure_name(C) : qualified_name;
 
-	string shared_ptr = "std::shared_ptr";
-	bool use_custom_shared = Config::get().use_custom_shared();
-	if( use_custom_shared )
-		shared_ptr = Config::get().custom_shared();
+	string holder_type = Config::get().holder_type();
 
+	string maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
 
-	string maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
-	//Check if the type is a custom shared pointer:
-	if( use_custom_shared && qualified_name.rfind(shared_ptr, 0) == 0 ) {
-		maybe_holder_type = "";
-	}
-	else if( is_inherited_from_enable_shared_from_this(C) ) {
-		maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
+	if( is_inherited_from_enable_shared_from_this(C) ) {
+		maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
 	}
 	else if( CXXDestructorDecl *d = C->getDestructor() ) {
 		if( d->getAccess() != AS_public ) maybe_holder_type = ", " + qualified_name + '*';
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -868,12 +868,25 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
 
 	string const include = relevant_include(C);
 
+	bool use_custom_shared = Config::get().use_custom_shared();
+	string const custom_shared = Config::get().custom_shared();
+
+	// Do not use custom_shared pointer if std:: class
+	if( qualified_name.rfind("std::", 0) == 0 ) use_custom_shared = false;
+
 	string c = "\t// Forward declaration for: " + qualified_name + " file:" + (include.size() ? include.substr(1, include.size() - 2) : "") + " line:" + line_number(C) + "\n";
 
 	string maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
-	if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
-	else if( CXXDestructorDecl *d = C->getDestructor() ) {
-		if( d->getAccess() != AS_public ) maybe_holder_type = ", " + qualified_name + '*';
+	if( use_custom_shared ) {
+		//Check if the type is a custom shared pointer:
+		if( qualified_name.rfind(custom_shared, 0) == 0 ) maybe_holder_type = "";
+		else maybe_holder_type = ", {}<{}>"_format(custom_shared, qualified_name);
+	}
+	else {
+		if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
+		else if( CXXDestructorDecl *d = C->getDestructor() ) {
+			if( d->getAccess() != AS_public ) maybe_holder_type = ", " + qualified_name + '*';
+		}
 	}
 
 	c += '\t' + R"(pybind11::class_<{}{}>({}, "{}");)"_format(qualified_name, maybe_holder_type, module_variable_name, python_class_name(C)) + "\n\n";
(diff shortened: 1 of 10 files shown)


source/class.cpp
```diff
@@ -868,12 +868,25 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
 
 	string const include = relevant_include(C);
 
+	bool use_custom_shared = Config::get().use_custom_shared();
+	string const custom_shared = Config::get().custom_shared();
+
+	// Do not use custom_shared pointer if std:: class
+	if( qualified_name.rfind("std::", 0) == 0 ) use_custom_shared = false;
```
lyskov, line 875: Why not? Could you please elaborate on this?

source/class.cpp
```diff
@@ -1172,12 +1185,24 @@ void ClassBinder::bind(Context &context)
 	string const trampoline_name = callback_structure_constructible ? callback_structure_name(C) : "";
 	string const binding_qualified_name = callback_structure_constructible ? callback_structure_name(C) : qualified_name;
 
+	bool use_custom_shared = Config::get().use_custom_shared();
+	string const custom_shared = Config::get().custom_shared();
+
+	// Do not use custom_shared pointer if std:: class
+	if( qualified_name.rfind("std::", 0) == 0 ) use_custom_shared = false;
```
lyskov, line 1192: Same as above: why should we exclude `std::` classes from using custom-shared-pointer-class?

source/context.cpp
```diff
@@ -429,7 +430,14 @@ void Context::generate(Config const &config)
 
 		if( i < binders.size() ) --i;
 
-		code = generate_include_directives(includes) + fmt::format(module_header, config.includes_code()) + prefix_code + "void " + function_name + module_function_suffix + "\n{\n" + code + "}\n";
+		string const custom_shared = Config::get().custom_shared();
+		bool use_custom_shared = Config::get().use_custom_shared();
+
+		string custom_shared_declare = "";
+		if( use_custom_shared )
+			custom_shared_declare = "PYBIND11_DECLARE_HOLDER_TYPE(T, "+custom_shared+"<T>)";
```
lyskov, line 438: Should we instead replace `PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>)` line with our custom shared line?

lyskov (commented):
Thank you for putting this together @kliegeois ! This looks good to me, a few thoughts:
- will you be open to also add simple test for this? If so please see https://github.com/RosettaCommons/binder/tree/master/test for list of example of how to do it (particularly cases with custom configs files).
- have you considered per-class specification of custom-share class? (i am not saying we should go this route, rather want to hear your take on that)
- please also see a few questions below.


Thanks,

lyskov:
Also, could you please add a few lines documenting this feature here: https://github.com/RosettaCommons/binder/blob/master/documentation/config.rst 
-- thanks,

kliegeois:
@lyskov thanks for your review! I will try to update the PR as soon as possible, probably next week.

kliegeois:
> * have you considered per-class specification of custom-share class? (i am not saying we should go this route, rather want to hear your take on that)

I did not consider this, it should be feasible to do this and to compile all the used custom-share classes (if more than one) when writing the PYBIND11_DECLARE_HOLDER_TYPE lines.

kliegeois:
@lyskov I updated the PR based on your comments and had a test. I created a second PR to update https://github.com/RosettaCommons/binder/blob/master/documentation/config.rst

Let me know if you need more modifications.

source/class.cpp
```diff
@@ -868,10 +868,22 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
...
+
 	string c = "\t// Forward declaration for: " + qualified_name + " file:" + (include.size() ? include.substr(1, include.size() - 2) : "") + " line:" + line_number(C) + "\n";
 
-	string maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
-	if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
+	string maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
+	//Check if the type is a custom shared pointer:
+	if( use_custom_shared && qualified_name.rfind(shared_ptr, 0) == 0 ) {
+		maybe_holder_type = "";
+	}
```
lyskov, line 883: Could you please elaborate what is going on here, why do we need this special case? Are we trying to restrict bindings of `custom_shared_ptr` class itself? If so then this check should rather go into https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L662

source/class.cpp
```diff
@@ -868,10 +868,22 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
 
 	string const include = relevant_include(C);
 
+	string shared_ptr = "std::shared_ptr";
+	bool use_custom_shared = Config::get().use_custom_shared();
+	if( use_custom_shared )
+		shared_ptr = Config::get().custom_shared();
```
lyskov, line 874: Optional refactoring:
```C++
bool use_custom_shared = Config::get().use_custom_shared();
string shared_ptr = use_custom_shared ? Config::get().custom_shared() : "std::shared_ptr";
```

source/class.cpp
```diff
@@ -1172,12 +1184,23 @@ void ClassBinder::bind(Context &context)
 	string const trampoline_name = callback_structure_constructible ? callback_structure_name(C) : "";
 	string const binding_qualified_name = callback_structure_constructible ? callback_structure_name(C) : qualified_name;
 
-	string maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name); // for now enable std::shared_ptr by default
-	if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
+	string shared_ptr = "std::shared_ptr";
+	bool use_custom_shared = Config::get().use_custom_shared();
+	if( use_custom_shared )
+		shared_ptr = Config::get().custom_shared();
```
lyskov, line 1190: It is second time we calculate this, perhaps adding dedicated function would be best to avoid code duplication? It might be best to put this into config.cpp 🤔
```C++
std::string shared_ptr_class_name() {
  return shared_ptr = Config::get().use_custom_shared() ? Config::get().custom_shared() : "std::shared_ptr"; 
}
```

source/class.cpp
```diff
@@ -1172,12 +1184,23 @@ void ClassBinder::bind(Context &context)
...
+
+
+	string maybe_holder_type = ", {}<{}>"_format(shared_ptr, qualified_name);
+	//Check if the type is a custom shared pointer:
+	if( use_custom_shared && qualified_name.rfind(shared_ptr, 0) == 0 ) {
+		maybe_holder_type = "";
+	}
```
lyskov, line 1197: Again, could you please elaborate whats going on here? Thanks,

source/config.hpp
```diff
@@ -50,6 +50,7 @@ class Config
 	string default_member_lvalue_reference_return_value_policy_ = "pybind11::return_value_policy::automatic";
 	string default_member_rvalue_reference_return_value_policy_ = "pybind11::return_value_policy::automatic";
 	string default_call_guard_ = "";
+	string custom_shared_ = "";
```
lyskov, line 53: Optional: It feel like this should have default `std::shared_ptr`, but then whole name of this option became a bit wrong, perhaps something like `holder_type` would be a bit more accurate, thoughts?

source/config.hpp
```diff
@@ -79,6 +80,9 @@ class Config
 	string const &default_member_rvalue_reference_return_value_policy() { return default_member_rvalue_reference_return_value_policy_; }
 	string const &default_call_guard() { return default_call_guard_; }
 
+	string const &custom_shared() const { return custom_shared_; }
+	bool use_custom_shared() const { return !custom_shared_.empty(); }
```
lyskov, line 84: Do we really need this flag?

test/T60.custom_shared.hpp
```diff
@@ -0,0 +1,43 @@
...
+/// @brief  Binder self-test file. Bindings of enum's functionality.
+/// @author Sergey Lyskov
+
+#ifndef _INCLUDED_T60_hpp_
+#define _INCLUDED_T60_hpp_
+
+#include <memory>
+
+template<typename T>
+using my_shared_ptr = std::shared_ptr<T>;
```
lyskov, line 19: +1 on this idea

source/class.cpp
kliegeois: Yes, what we want to avoid here is to create a `my_shared_ptr<my_shared_ptr<T>>`. If the `qualified_name` is already a custom shared pointer, we should not add a holder type associated to a custom shared pointer to a custom shared pointer.
Yes, I agree, it would be cleaner to have it with the other checks in https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L662 , sorry I was unaware of those.

source/class.cpp
kliegeois: Yes, it is cleaner.

source/class.cpp
kliegeois: Same as above and same comment w.r.t https://github.com/RosettaCommons/binder/blob/master/source/type.cpp#L662

source/config.hpp
kliegeois: Yes and that would simplify this:
```
std::string shared_ptr_class_name() {
  return shared_ptr = Config::get().use_custom_shared() ? Config::get().custom_shared() : "std::shared_ptr"; 
}
```
Instead, we could just use:
`Config::get().holder_type()`
with 
`string const &holder_type() const { return holder_type_; }`

Actually, with all of the above comments, we can even get rid of the function `use_custom_shared()`.

source/config.hpp
kliegeois: No, we don't if we do as discussed above.

kliegeois:
I am going to update the PR, thanks @lyskov for the review!

lyskov:
@kliegeois bumping this PR

kliegeois:
@lyskov , I am currently out of the office but I am planning to work on this by the end of the month.

kliegeois:
@lyskov the PR is updated as discussed in the comments. The tests passed on my machine.

kliegeois:
We can close https://github.com/RosettaCommons/binder/pull/219 as it is included in this PR now.

source/class.cpp
```diff
@@ -868,10 +868,15 @@ string bind_forward_declaration(CXXRecordDecl const *C, Context &context)
 
 	string const include = relevant_include(C);
 
+	string holder_type = Config::get().holder_type();
+
 	string c = "\t// Forward declaration for: " + qualified_name + " file:" + (include.size() ? include.substr(1, include.size() - 2) : "") + " line:" + line_number(C) + "\n";
 
-	string maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
-	if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
+	string maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
```
lyskov, line 875: Optional: looks like original code have duplication, this could be refactored as `string maybe_holder_type;`

source/class.cpp
```diff
@@ -1172,12 +1177,16 @@ void ClassBinder::bind(Context &context)
 	string const trampoline_name = callback_structure_constructible ? callback_structure_name(C) : "";
 	string const binding_qualified_name = callback_structure_constructible ? callback_structure_name(C) : qualified_name;
 
-	string maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name); // for now enable std::shared_ptr by default
-	if( is_inherited_from_enable_shared_from_this(C) ) maybe_holder_type = ", std::shared_ptr<{}>"_format(qualified_name);
+	string holder_type = Config::get().holder_type();
+
+	string maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);
```
lyskov, line 1182: Optional: looks like original code have duplication, this could be refactored as `string maybe_holder_type;`

lyskov (approved):
Looks good to me, - thank you for putting this together @kliegeois ! I have two minor and very optional comments unless you want to merge this as-is - please let me know. Thanks,

source/class.cpp
kliegeois: Are we sure that it will be always correct to replace:
`string maybe_holder_type = ", {}<{}>"_format(holder_type, qualified_name);`
with
`string maybe_holder_type;`
?

The following logic that sets `maybe_holder_type` is 
```
if {
}
else if{
  if
}
```

I think that I would be in favor of leaving the logic as in the original code and merge it as it is.
Would you be fine with this?

kliegeois:
Thanks for the reviews and comments @lyskov !

source/class.cpp
lyskov: Yes, let's leave the old logic for now and i will refactor this a bit later.

PR #218, opened 2022-09-01
Removed too strict check for binding request, allows template classes to be easily bound

mattangus:
This PR makes using templates easier. For example before this change the following would generate no output. After this PR it will now bind the template class to `TestClass_float_t` as well as `TestClass_float_t.data`

```c++
namespace test
{
  template<typename T>
  class TestClass {
  public:
    T data;
  };

  template class TestClass<float>;

} // namespace test
```

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -333,7 +333,6 @@ bool is_bindable_raw(clang::CXXRecordDecl const *C)
 /// check if user requested binding for the given declaration
 bool is_binding_requested(clang::CXXRecordDecl const *C, Config const &config)
 {
-	if( dyn_cast<ClassTemplateSpecializationDecl>(C) ) return false;
 	bool bind = config.is_class_binding_requested(standard_name(C->getQualifiedNameAsString())) or config.is_class_binding_requested(class_qualified_name(C)) or
 				config.is_namespace_binding_requested(namespace_from_named_decl(C));
 	for( auto &t : get_type_dependencies(C) ) bind &= !is_skipping_requested(t, config);


lyskov (commented):
Interesting, - thank you for letting me know about this! Looks like indeed current checks is indeed a bit too restricting. Though i can not merge this as-is as it leads to unbindable code in my project, so it looks like some extra checks will be needed. Let me investigate this and i will get back to you.

lyskov:
Ok, i played with this and i can not make this work for complicated template cases. Here is the example that i have in mind, consider following code (below). It will compile with current template-binding-logic but fail with proposed patch. And in projects with moderate template usage there will be much more 'fail' cases then passed one so requiring disabling bindings for all such cases would be quite a task. Thoughts?
```C++
struct A {
private:
	bool operator==(A const &) { return false;};
};


template<typename T>
class B {
public:
	bool operator==(T const &) { T t; return t==t;};
};

template<typename T>
class TestClass {
public:
    T data;

	bool foo(B<T>) {}

};

template class TestClass<float>;
template class TestClass<A>;
```

mattangus:
I was able to generate bindings with the snippet you posted. I'm generating the following:

```c++
void bind_TestClass(std::function< pybind11::module &(std::string const &namespace_) > &M)
{
	{ // test::A file:TestClass.hpp line:5
		pybind11::class_<test::A, std::shared_ptr<test::A>> cl(M("test"), "A", "");
		cl.def( pybind11::init( [](){ return new test::A(); } ) );
	}
	{ // test::B file:TestClass.hpp line:12
		pybind11::class_<test::B<float>, std::shared_ptr<test::B<float>>> cl(M("test"), "B_float_t", "");
		cl.def( pybind11::init( [](){ return new test::B<float>(); } ) );
		cl.def("__eq__", (bool (test::B<float>::*)(const float &)) &test::B<float>::operator==, "C++: test::B<float>::operator==(const float &) --> bool", pybind11::arg(""));
	}
	{ // test::B file:TestClass.hpp line:12
		pybind11::class_<test::B<test::A>, std::shared_ptr<test::B<test::A>>> cl(M("test"), "B_test_A_t", "");
		cl.def( pybind11::init( [](){ return new test::B<test::A>(); } ) );
		cl.def("__eq__", (bool (test::B<test::A>::*)(const struct test::A &)) &test::B<test::A>::operator==, "C++: test::B<test::A>::operator==(const struct test::A &) --> bool", pybind11::arg(""));
	}
	{ // test::TestClass file:TestClass.hpp line:31
		pybind11::class_<test::TestClass<float>, std::shared_ptr<test::TestClass<float>>> cl(M("test"), "TestClass_float_t", "");
		cl.def( pybind11::init( [](){ return new test::TestClass<float>(); } ) );
		cl.def_readwrite("data", &test::TestClass<float>::data);
		cl.def("foo", (bool (test::TestClass<float>::*)(class test::B<float>)) &test::TestClass<float>::foo, "C++: test::TestClass<float>::foo(class test::B<float>) --> bool", pybind11::arg(""));
	}
	{ // test::TestClass file:TestClass.hpp line:32
		pybind11::class_<test::TestClass<test::A>, std::shared_ptr<test::TestClass<test::A>>> cl(M("test"), "TestClass_test_A_t", "");
		cl.def( pybind11::init( [](){ return new test::TestClass<test::A>(); } ) );
		cl.def_readwrite("data", &test::TestClass<test::A>::data);
		cl.def("foo", (bool (test::TestClass<test::A>::*)(class test::B<struct test::A>)) &test::TestClass<test::A>::foo, "C++: test::TestClass<test::A>::foo(class test::B<struct test::A>) --> bool", pybind11::arg(""));
	}
}
```

This doesn't seem correct, but it is generating code for me.

lyskov:
-- sorry if my previous post was a bit cryptic! - indeed Binder will be able to generate bindings for mentioned test case but generate code will be be compilable (due to `A::operator==` being private).

I meditated a bit more on this and i think it seems reasonable that there will be projects that will benefit from template specialization being bound and not bound by default. So making command line option to control this behavior would be a best approach. I will push related patch soon and post here when it is in master.

lyskov:
@mattangus  i pushed patch that added `--bind-class-template-specialization` option (enable by-default) into master. Hope this helps. (i am also closing this PR since it is now redundant)

lyskov:
@mattangus  i pushed patch that added `--bind-class-template-specialization` option (enable by-default) into master, please see 
https://github.com/RosettaCommons/binder/commit/156eef05fb867a132fa17606059ad3d9a14a5693 . -Hope this helps. 
(i am also closing this PR since it is now redundant)

PR #222, opened 2022-09-16
Gp

kkrissian-gpfw:
Add '+buffer_protocol classname' configuration option for a given class, it will allow using the def_buffer() method of pybind11 to define the buffer protocol for the class.


lyskov (commented):
Thank you for putting this together @kkrissian-gpfw ! - this PR looks good to me, i will merge it as soon as i see green-light from CI.

lyskov (approved):

Issue #221, opened 2022-09-15
Binding of code using `pybind11` objects.


zwimer:
TLDR: Allow functions using `pybind11::dict` without generating bindings for `pybind11::dict` since pybind11 does not need custom bindings for its own objects.

---

Consider:
```C++
namespace Test { inline std::string foo(pybind11::dict& d) { return pybind11::repr(d); } }
```

If I want to generate bindings for `+namespace Test` binder generates many files worth of code because it generates bindings for `pybind11::dict` and dependencies (the log is below). But this shouldn't need to be done since when compiling with the pybind11 headers pybind11 will know exactly what `pybind11::dict` is.

If I try to do `-class pybind11::dict` or something like that, the code for the `foo` function simply will not generate. Also, a lot of the code is put into `unknown/` rather than the `pybind11/` output directories, so skipping that directory is not sufficient for removing the unwanted code.

Is there a good way to either skip generating code for `pybind11::dict` or to simply remove it from the output directory after the fact?

---

```
Process input file /in/include.hpp
Read config from file /in/a.conf
Generate bindings, pass 1...
Binding: class std::basic_string<char> Test::foo(class pybind11::dict &)
Generate bindings, pass 2...
Binding: pybind11::dict
Generate bindings, pass 3...
Binding: pybind11::handle
Binding: pybind11::object
Binding: pybind11::detail::generic_iterator<pybind11::detail::iterator_policies::dict_readonly>
Binding: pybind11::detail::iterator_policies::dict_readonly
Binding: pybind11::str
Binding: pybind11::bytes
Generate bindings, pass 4...
Binding: _object
Binding: pybind11::detail::object_api<pybind11::handle>
Binding: pybind11::detail::args_proxy
Binding: pybind11::iterator
Generate bindings, pass 5...
Binding: pybind11::detail::accessor<pybind11::detail::accessor_policies::obj_attr>
Binding: pybind11::detail::accessor<pybind11::detail::accessor_policies::str_attr>
Binding: pybind11::detail::accessor<pybind11::detail::accessor_policies::generic_item>
Binding: pybind11::detail::pyobject_tag
Binding: pybind11::detail::object_api<pybind11::detail::accessor<pybind11::detail::accessor_policies::str_attr>>
Binding: pybind11::detail::object_api<pybind11::detail::accessor<pybind11::detail::accessor_policies::generic_item>>
[...]

PR #223, opened 2022-09-20
STL flag to enable STL bindings

zwimer:
This PR labels pybind11 objects as builtins. For example `pybind11::dict` and `pybind11::list`.

### DIFF
--- a/source/type.cpp
+++ b/source/type.cpp
@@ -746,7 +746,7 @@ bool is_python_builtin(NamedDecl const *C)
 	if (Config::get().python_builtins.count(name) || known_builtin.count(name))
 		return true;
 	// STL
-	if (O_stl && Config::get().stl_builtin.count(name))
+	if (O_include_pybind11_stl && Config::get().stl_builtin.count(name))
 		return true;
 
 	return false;
--- a/source/type.cpp
+++ b/source/type.cpp
@@ -653,7 +653,7 @@ bool is_python_builtin(NamedDecl const *C)
 	string name = standard_name(C->getQualifiedNameAsString());
 	// if( begins_with(name, "class ") ) name = name.substr(6); // len("class ")
 
-	static std::vector<string> const known_builtin = {
+	static std::set<string> const known_builtin = {
 		//"std::nullptr_t", "nullptr_t",
 		"std::basic_string", "std::initializer_list", "std::__1::basic_string",
 
@@ -683,8 +683,6 @@ bool is_python_builtin(NamedDecl const *C)
 		"std::__hash_value_type",
 
 		"std::function", "std::complex",
-
-		"std::optional",
 
 		// pybind11 types
 		// https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html
@@ -716,7 +714,6 @@ bool is_python_builtin(NamedDecl const *C)
 		"pybind11::frozenset",
 		"pybind11::set",
 		"pybind11::function",
-		"pybind11::function",
 		"pybind11::cpp_function",
 		"pybind11::staticmethod",
 		"pybind11::buffer",
@@ -725,17 +722,32 @@ bool is_python_builtin(NamedDecl const *C)
 		"pybind11::array_t",
 	};
 
+	static std::set<string> const stl_builtin = {
+		"std::vector",
+		"std::deque",
+		"std::list",
+		"std::array",
+		"std::valarray",
+		"std::set",
+		"std::unordered_set",
+		"std::map",
+		"std::unordered_map",
+		// C++14
+		"std::experimental::optional",
+		// C++17
+		"std::optional",
+		"std::variant",
+	};
+
 	// Not builtin's
-	for ( const auto &k : Config::get().not_python_builtins ) {
-		if ( name == k ) return false;
-	}
+	if (Config::get().not_python_builtins.count(name))
+		return false;
 	// Builtins
-	for( const auto &k : known_builtin ) {
-		if( name == k ) return true;
-	}
-	for( const auto &k : Config::get().python_builtins ) {
-		if( name == k ) return true;
-	}
+	if (Config::get().python_builtins.count(name) || known_builtin.count(name))
+		return true;
+	// STL
+	if (O_stl && Config::get().stl_builtin.count(name))
+		return true;
 
 	return false;
 }
--- a/source/type.cpp
+++ b/source/type.cpp
@@ -746,7 +746,7 @@ bool is_python_builtin(NamedDecl const *C)
 	if (Config::get().python_builtins.count(name) || known_builtin.count(name))
 		return true;
 	// STL
-	if (O_include_pybind11_stl && Config::get().stl_builtin.count(name))
+	if (O_include_pybind11_stl && stl_builtin.count(name))
 		return true;
 
 	return false;
--- a/test/T43.stl.pybind11_include_stl.cli
+++ b/test/T43.stl.pybind11_include_stl.cli
@@ -1,0 +1,1 @@
+--include-pybind11-stl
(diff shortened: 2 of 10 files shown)


zwimer:
Related to: https://github.com/RosettaCommons/binder/issues/221

lyskov (commented):
I like the idea of giving user control on how to handle STL types. However the proposed solution give only all-or-nothing which might not be the optimal way for most projects. How about if we create new config option `+python_builtin <type>` which will allow user to fully customize such behavior. Thoughts?

zwimer:
I like the idea of both, but alone yours is a more flexible and better idea as it would also allow users to do `python_builtin abc` for any custom bindings for `abc` they've made.

I do not know how to implement this, however, which is why I made this PR as a stop-gap.

zwimer:
@lyskov I made a PR with the feature you mentioned. I still think this PR should also be merged however. `#include <pybind11/stl.h>` feels common enough to have this 'set' of `+python_builtin` config options just be a flag. Users generally should _not_ have to manually do a `+python_builtin` if both https://github.com/RosettaCommons/binder/pull/230 and https://github.com/RosettaCommons/binder/pull/224 are merged I think, though this gives them that option should they desire it.

The only change I would say we should add if we merge al of these, is the docs should reflect that `+python_builtin` will (or will not depending on how we merge these) supersede `--stl`.

zwimer:
We may also want to have this auto-enabled if `+include <pybind11/stl.h>` is in the config file?

lyskov:
> We may also want to have this auto-enabled if `+include <pybind11/stl.h>` is in the config file?

-- yes, this seems reasonable.

lyskov:
> @lyskov I made a PR with the feature you mentioned. I still think this PR should also be merged however. `#include <pybind11/stl.h>` feels common enough to have this 'set' of `+python_builtin` config options just be a flag. Users generally should _not_ have to manually do a `+python_builtin` if both #230 and #224 are merged I think, though this gives them that option should they desire it.

-- I agree, - it certainly looks attractive to have one option as turn-key solution so i like this idea in general. To merge this i see two things that will need to updated: 

[a] lets change the name of command-line-option to something more descriptive. Current `--stl` is rather confusing since it does not reflect what exactly will be happening. I thinking maybe `--use-pybind11-auto-conversion` or something, - thoughts?

[b] lets also update docs to clearly reflect upside and downside of using this option (particularly, clearly stating performance penalties and that when this option is used then modification in-place of STL data structures will not work as expected. 

[c] (optional) add simple test

zwimer:
[a] How about `--include-pybind11-stl` referencing `#include <pybind11/stl.h>` ?

[b] Sounds good to me.

[c] I'll give that a shot

One thing we maybe want this to do is `#include <pybind11/stl.h>` in the output files it produces when this flag is enabled? If so, any advice on how to change binder to do this would be useful.

lyskov:
> [a] How about `--include-pybind11-stl` referencing `#include <pybind11/stl.h>` ?

-- good idea! Lets go with `--include-pybind11-stl`.

> One thing we maybe want this to do is `#include <pybind11/stl.h>` in the output files it produces when this flag is enabled? If so, any advice on how to change binder to do this would be useful.

  -- i though i replied to this in previous post: - yes this sounds like a good approach to me. To implement this i think simplest approach would be to conditionally modify module-header template (https://github.com/RosettaCommons/binder/blob/master/source/context.cpp#L130). This could be done at the end of this function https://github.com/RosettaCommons/binder/blob/master/source/config.cpp#L362 by conditionally including pybind11/stl.h based on state of command line option above.

Thanks,

zwimer:
I rebased, addressed the above, update documentation to specify how `python_builtin` interacts with this by noting that `-python_builtin` always overrides everything, fixed a few typos. I tested the `config.rst` via `rst2pdf`.

zwimer:
For the test I can do: `cat T43.stl.pybind11_include_stl.hpp`:
```C++
#ifndef _INCLUDED_T40_pybind11_include_stl_hpp_
#define _INCLUDED_T40_pybind11_include_stl_hpp_

#include <vector>
#include <deque>
#include <list>
#include <array>
#include <valarray>
#include <set>
#include <unordered_set>
#include <map>
#include <unordered_map>

#define TEST_ID(TYPE, ...) inline std::TYPE<__VA_ARGS__> test_##TYPE(std::TYPE<__VA_ARGS__> & a) { return a; }

namespace TestSTLInc {

	TEST_ID(vector, int)
	TEST_ID(deque, int)
	TEST_ID(list, int)
	TEST_ID(array, int, 5)
	TEST_ID(valarray, int)
	TEST_ID(set, int)
	TEST_ID(unordered_set, int)
	TEST_ID(map, int, int)
	TEST_ID(unordered_map, int, int)

}

#undef TEST_ID

#endif // _INCLUDED_T40_pybind11_include_stl_hpp_
```

lyskov:
> For the test I can do: `cat T43.stl.pybind11_include_stl.hpp`:
> 
> ```c++
>... 
> 	TEST_ID(vector, int)
>...
> ```

-- that would be great! Optional: I would personally avoid using macros and instead do something like:
```c++
void test_pybind_stl(std::vector, std::deque, std::list, MORE-TYPES-HERE) {}
```

Thanks,

zwimer:
@lyskov What I don't know is how to actually add this test. My thoughts:
1. Run binder locally (possibly with `--single-file`?) to generate a `.ref`?
2. CMake has to be informed that this test requires a different CLI args with `--include-pybind11-stl`
3. A `.conf` file that has `+TestSTLInc`

I'm not sure if a `.ref` is required or not; one was not edited when we added the test cases for the `+python_builtin` PR; there aren't really docs I can refer to about adding test cases.

The proposed test case above tries to create a function for each of the STL containers that the `--pybind11-include-stl` flag adds. What it does not have:
```C++
// C++14
"std::experimental::optional",
// C++17
"std::optional",
"std::variant"
```
I did not add these because I'm not sure what version of C++ your tests are using, and obviously `--std=c++11` would not play well with `std::optional` when trying to compile

source/type.cpp
```diff
@@ -653,7 +653,7 @@ bool is_python_builtin(NamedDecl const *C)
 	string name = standard_name(C->getQualifiedNameAsString());
 	// if( begins_with(name, "class ") ) name = name.substr(6); // len("class ")
 
-	static std::vector<string> const known_builtin = {
+	static std::set<string> const known_builtin = {
```
lyskov, line 656: 👍

lyskov (dismissed):
Looks good to me! Please ping me when you add test (or decide against it) and i will merge this, - thank you for putting this together @zwimer !

zwimer:
@lyskov I edited the `cmake` so that if there is a `.cli` file in the tests dir it will edit the CLI arguments used for that test's binder invocation. I also added a `.ref` (I just copied the output file that `ctest` generated assuming that was fine).

lyskov:
re ref files: they are sensitive to compiler/libstdc++ version so it best to include them (i will generate them later in CentOS-7 env).

lyskov (approved):
Looks good to me, - thank you for adding this @zwimer !

test/T43.stl.pybind11_include_stl.hpp
```diff
@@ -0,0 +1,40 @@
...
+	std::set<int>,
+	std::unordered_set<int>,
+	std::map<int,int>,
+	std::unordered_map<int,int>
+#if ((defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L)
+    // C++17
+	,
+	std::optional<int>,
+	std::variant<int, bool>
+#endif
```
lyskov, line 37: +1

test/T43.stl.pybind11_include_stl.cli
```diff
@@ -0,0 +1 @@
+--include-pybind11-stl
```
lyskov, line 1: good idea on adding `.cli` file!

Issue #232, opened 2022-10-07
`__div__` does not exist in python 3.x, only 2.x. Python3 uses either `__truediv__` or `__floordiv__`


zwimer:
Binder currently maps `operator/` to `__div__`; but that magic function doesn't exist in python3, which I assume is what binder generates for given that python2 was end of life years ago.

You can see the bug here:
https://github.com/RosettaCommons/binder/blob/7c28c46af1008119237b327d9b1fbfcb23d844a8/source/function.cpp#L47

This is also true for `__idiv__`.

lyskov:
Thank you for letting me know @zwimer ! This is now fixed in 5f0a7bd

zwimer:
@lyskov This doesn't fix `__itruediv__` for `/=`

PR #233, opened 2022-10-07
fix division binding issue

zwimer:
Fixes https://github.com/RosettaCommons/binder/issues/232

### DIFF
--- a/source/function.cpp
+++ b/source/function.cpp
@@ -44,12 +44,12 @@ static std::map<string, string > const cpp_python_operator_map{
 	{"operator+", "__add__"}, //
 	{"operator-", "__sub__"}, //
 	{"operator*", "__mul__"}, //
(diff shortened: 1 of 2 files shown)


zwimer:
I chose to switch to true division, instead of floor division, mostly because the signs are the same. I.e. python does `3 // 2` for floor division and `3 / 2` for true division. Since these are for use defined classes we could
1. Do truediv (I did this)
2. Do floordiv
3. Do both

lyskov:
Thank you for PR @zwimer ! Unfortunately i saw this only after i pushed my fix to master so i will just close this. (all these changes now in master)

zwimer:
@lyskov This also fixes `__itruediv__` for `/=`.

zwimer:
Oh, I see they are both updated in master now!

Issue #227, opened 2022-09-20
Operator bindings undocumented


zwimer:
Which operators bind, are ignored, and only bind under certain circumstances is undocumented.

zwimer:
I will make a PR to the docs about this

zwimer:
It'll be the same PR as for this issue: https://github.com/RosettaCommons/binder/issues/220

zwimer:
Fixed via https://github.com/RosettaCommons/binder/pull/229

PR #234, opened 2022-10-07
Add mod support

zwimer:
Support for `operator%` and `operator%=`

It should be worth noting that `%` in C++ and `%` in python are not exact analogs for each other. For example `-7%3` gives `-1` in `C++` and `2` in python. Both are mathematically valid, but python typically also requires the result be positive, it's more of a division remainder. These would only be binding for custom classes / definitions though so I think this is ok to gloss over.


zwimer:
If merged, documentation would be need to updated in: https://github.com/RosettaCommons/binder/issues/227
Or if that gets merged first, this would have to update docs in this PR

lyskov (approved):
Looks good to me, - thank you for putting this together @zwimer !

Issue #226, opened 2022-09-20
Post/Pre increment/decrement operator confusion


zwimer:
Both `++a` and `a++` map to `plus_plus`, though as one the differentiation between them still exists via different overrides.

That is `++a` maps to `plus_plus()` and `a++` to `plus_plus(0)`; which makes sense given how the operators are defined in C++.

A feature add could be changing `plus_plus` to `post_increment` / `pre_increment` depending on which is detected. All of this is true with `--` too.

zwimer:
This was documented via https://github.com/RosettaCommons/binder/pull/229

The rest of this issue is just a feature request at this point.

zwimer:
Feature add PR: https://github.com/RosettaCommons/binder/pull/238

PR #238, opened 2022-10-12
Pre post increment ops renamed

zwimer:
This is a feature add that also fixes the same bugs that https://github.com/RosettaCommons/binder/pull/237 fixes (as the code introduced in https://github.com/RosettaCommons/binder/pull/237 made it easy to add this feature). It addresses https://github.com/RosettaCommons/binder/issues/226

### DIFF
--- a/source/function.cpp
+++ b/source/function.cpp
@@ -40,40 +40,54 @@ using namespace fmt::literals;
 
 namespace binder {
 
-static std::map<string, string > const cpp_python_operator_map{
-	{"operator+", "__add__"}, //
-	{"operator-", "__sub__"}, //
-	{"operator*", "__mul__"}, //
-	{"operator/", "__truediv__"}, //
-	{"operator%", "__mod__"}, //
-	{"operator~", "__invert__"}, //
-	{"operator|", "__or__"}, //
-	{"operator&", "__and__"}, //
-	{"operator^", "__xor__"}, //
-	{"operator<<", "__lshift__"}, //
-	{"operator>>", "__rshift__"}, //
+// Return the python operator that maps to the C++ operator; returns "" if no mapping exists
+string cpp_python_operator(const FunctionDecl & F) {
+	static std::map<string, vector<string>> const m {
+		{"operator+", {"__pos__", "__add__"}}, //
+		{"operator-", {"__neg__", "__sub__"}}, //
+		{"operator*", {"dereference", "__mul__"}}, //
+		{"operator/", {"__truediv__"}}, //
+		{"operator%", {"__mod__"}}, //
+		{"operator~", {"__invert__"}}, //
+		{"operator|", {"__or__"}}, //
+		{"operator&", {"__and__"}}, //
+		{"operator^", {"__xor__"}}, //
+		{"operator<<", {"__lshift__"}}, //
+		{"operator>>", {"__rshift__"}}, //
 
-	{"operator+=", "__iadd__"}, //
-	{"operator-=", "__isub__"}, //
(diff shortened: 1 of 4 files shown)


zwimer:
As mentioned in https://github.com/RosettaCommons/binder/pull/237 there are a few possible options we can do:

1. Close this and post/pre increment both named `plus_plus`, with post requiring a integer dummy argument (This makes it harder to tell `++a` from `a++` apart)
2. Merge this and rename define `pre_increment()` and `post_increment(int)` (This is not backwards compatible)
3. Improve this and define `pre_increment()` and `post_increment()` (This would remove the ability to control that dummy integer in the rare case someone decided to use it for some reason)
4. Do either 2 or 3, but also keep `plus_plus` for compatibility (This might add overhead)
5. Add a CLI flag to decide between options

zwimer:
Should I do anything to get this PR moving?

source/function.cpp
```diff
@@ -73,8 +73,8 @@ string cpp_python_operator(const FunctionDecl & F) {
 		{"operator!=", {"__ne__"}}, //
 		{"operator[]", {"__getitem__"}}, //
 		{"operator=", {"assign"}}, //
-		{"operator++", {"plus_plus"}}, //
-		{"operator--", {"minus_minus"}}, //
+		{"operator++", {"pre_increment", "pre_increment"}}, //
```
lyskov, line 76: probably `{"operator++", {"pre_increment", "post_increment"}},` was intended

lyskov (commented):
Sorry if my previous comment on was a bit cryptic. The main issue with merging this is backward compatibility with already generate code. However i just checked and it looks like there is only a few cases where these was uses in my project so i am open on merging this. Please see minor comment below that needs to be address before this is merged. Thanks,

lyskov (approved):
LGTM, - thank you for putting this together @zwimer !

lyskov:
@aleaverfay Hi Andrew! Sorry to hear that these changes broke your code! These change have not break any of PyRosetta unit tests so I am guessing that it could be something specific to area of code you are working with. Never the less let's see if we can find a good solution that works for you!

I am a bit skeptical about "double bindings" since it will just prolong the issue (and then we will have one more API function to take care of) (and currently Binder does not have such functionality). The code that got broken, - is that something that you will be OK/easy/have-permission to modify? If so, have you considered the following workaround which should have ~zero performance impact:

During import phase, we can check if particular class already have `plus_plus` defined and if not just link its implementation to `pre_increment`/`post_increment`, like this:

```Python
...
import aaaa
if not hasattr(aaaa.A, 'plus_plus'): aaaa.A.plus_plus = aaaa.A.pre_increment
```

-- will something like this work for you? Thanks,

lyskov:
@aleaverfay Hi Andrew! Sorry to hear that these changes broke your code! These change have not break any of PyRosetta unit tests (and I have not heard so far from other users regarding this change) so I am guessing that it could be something specific to area of code you are working with. Never the less let's see if we can find a good solution that works for you!

I am a bit skeptical about "double bindings" since it will just prolong the issue (and then we will have one more API function to take care of) (and currently Binder does not have such functionality). The code that got broken, - is that something that you will be OK/easy/have-permission to modify? If so, have you considered the following workaround which should have ~zero performance impact:

During import phase, we can check if particular class already have `plus_plus` defined and if not just link its implementation to `pre_increment`/`post_increment`, like this:

```Python
...
import aaaa
if not hasattr(aaaa.A, 'plus_plus'): aaaa.A.plus_plus = aaaa.A.pre_increment
```

-- will something like this work for you? Thanks,

PR #240, opened 2022-10-14
The documentation has a bad reference

zwimer:
This should hopefully fix
> To statically compile binder, see [With Docker](https://cppbinder.readthedocs.io/en/latest/install.html#building-static).

With
> To statically compile binder, see [Building Statically (Linux only)](https://cppbinder.readthedocs.io/en/latest/install.html#building-static).

### DIFF
--- a/documentation/install.rst
+++ b/documentation/install.rst
@@ -209,13 +209,13 @@ If ``binder`` was build with some older versions of LLVM, one could also set the
 
 
 
-.. _building-static:
-
 With Docker
 ***********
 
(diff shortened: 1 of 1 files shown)


zwimer:
My `rst` viewer doesn't seem to recognize the `:ref:` syntax used so I can't entirely test this, but it would seem I just moved the marker before the proper section so I think it should be correct

lyskov (approved):
LGTM, - thank you @zwimer !

Issue #254, opened 2023-01-24
Is there a way to exclude an enum?


andriish:
Hi @lyskov,

it seems there is no way to exclude an enum from the bindings. Moreover, the `std::_` enums are binded. Or am I missing some options that can prevent that?

Andrii

andriish:
Oh, seems to be a duplicate of https://github.com/RosettaCommons/binder/issues/166.
Anyway.

lyskov:
@andriish i just pushed the patch that adds `enum` config option which will allow to include/exclude enum's. Hope this helps!

PR #255, opened 2023-01-24
Remove the banned symbols from enum bindings. Note: the banned symbols function notes enums.

andriish:
Remove the banned symbols from enum bindings. Note: the banned symbols function notes enums.
Partial solution for #254


lyskov (approved):
LGTM!

PR #257, opened 2023-04-26
Add option to skip binding a field

jlblancoc:
This commit is based on a larger commit by @charbeljc in https://github.com/RosettaCommons/binder/pull/160 but manually cherry-picking just the "-field" feature.

### DIFF
--- a/source/config.cpp
+++ b/source/config.cpp
@@ -57,6 +57,7 @@ void Config::read(string const &file_name)
 	string const _namespace_{"namespace"};
 	string const _function_{"function"};
 	string const _class_{"class"};
+	string const _field_{"field"};
 	string const _enum_{"enum"};
 
 	string const _python_builtin_{"python_builtin"};
@@ -206,6 +207,11 @@ void Config::read(string const &file_name)
 			if( bind ) {
 				auto binder_function = split_in_two(name, "Invalid line for add_on_binder_for_namespace specification! Must be: name_of_type + <space or tab> + name_of_binder. Got: " + line);
 				add_on_binder_for_namespaces_[binder_function.first] = binder_function.second;
+			}
+		} else if ( token == _field_ ) {
+
+			if (!bind) {
+				fields_to_skip.push_back(name_without_spaces);
 			}
 		}
 		else if( token == _custom_shared_ ) holder_type_ = name_without_spaces;
@@ -359,6 +365,19 @@ bool Config::is_class_skipping_requested(string const &class__) const
 		// outs() << "Skipping: " << class_ << "\n";
 		return true;
 	}
+
+	return false;
+}
+
+
+bool Config::is_field_skipping_requested(string const &field_) const
+{
+	string field {field_};
+	field.erase(std::remove(field.begin(), field.end(), ' '), field.end());
+
+	auto bind = std::find(fields_to_skip.begin(), fields_to_skip.end(), field);
+
+	if( bind != fields_to_skip.end() ) return true;
 
 	return false;
 }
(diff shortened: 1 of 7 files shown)


source/config.cpp
```diff
@@ -364,6 +370,19 @@ bool Config::is_class_skipping_requested(string const &class__) const
 }
 
 
+bool Config::is_field_skipping_requested(string const &field_) const
+{
+	string field {field_};
+	field.erase(std::remove(field.begin(), field.end(), ' '), field.end());
```
lyskov, line 376: Are you sure this is needed? At line 214 https://github.com/RosettaCommons/binder/pull/257/files#diff-72f8d59ab463236c54ef8c3a2c8fb31c72efad5003751b3e342c868079de19baR214 we added fields name without namespaces so it looks like it this line will not do anything. I am brining this up due to performance consideration: in large projects this check might be called numerous times.

lyskov (changes requested):
Thank you for implementing this @jlblancoc , - looks very useful! I adding minor comment on current code but otherwise it looks good and i will be happy to merge it. However, it would be great if we can also add minimal test for this functionality. Would you be interested in doing it? If yes please see https://github.com/RosettaCommons/binder/tree/master/test and particularly check tests with custom config files like https://github.com/RosettaCommons/binder/blob/master/test/T80.custom_trampoline.config. If that is too much then we can merge this as-is (after addressing comment below). Thanks,

jlblancoc:
Sure! One doubt: what version of clang-format are you using to format the files? I tried clang-format-11 and clang-format-15 and both changes much more than just my changes...

lyskov:
@jlblancoc re clang-format version: unless i am mistaken it was v13

jlblancoc:
Hi @lyskov ! It's done now: 
- You were right, the space trimming code wasn't needed. It's removed.
- Added a unit test. See https://github.com/RosettaCommons/binder/pull/257/commits/630cf4435ec0ab96dd5848017078a1001db3ef90

There are other weird things I noticed in the way, but kept changes minimal in this PR anyway, please check them for subsequent commits?: 
- test/self-test.py only works with python2.7. I had to manually change a couple of lines to make it work with python3.
- no clang-format version seems to be actually applied to existing sources (some of them). Please, check if running "clang-format-XX -i FILE" on all sources make changes for you too, and commit them so it's easier to keep the formatting for contributors  :-) 
- It's confusing why only part of the test/*.hpp files are handled at the `cmake` + `make test` level, but then there is an apparently independent way to run tests via "build-and-run-tests.py" (??).

That said... congrats for the *huge* project, it literally made my life easier :-)

lyskov (approved):
LGTM!

lyskov:
Thank you for your feedback and for adding test for this @jlblancoc !

> * test/self-test.py only works with python2.7. I had to manually change a couple of lines to make it work with python3.

-- yes, thank you for brining this up! I am aware of this and will escalate update to Python-3.

> * no clang-format version seems to be actually applied to existing sources (some of them). 

-- thank you for letting me know! Looks like quite a few changes got accumulated since last clang format run, - i will looks this up.

> * It's confusing why only part of the test/*.hpp files are handled at the `cmake` + `make test` level, but then there is an apparently independent way to run tests via "build-and-run-tests.py" (??).

-- cmake tests does not compare test results (only makes sure that results is compilable) and they also require system-wide install of LLVM which is not always practical during development.

Again, thank you for your PR and feedback @jlblancoc ! I will merge this shortly.

jlblancoc:
Got it. Thanks you!

Issue #258, opened 2023-05-11
Non-portable generated code for `size_t`


jlblancoc:
## Problem

```cpp
// C++ API to be wrapped:
size_t foo();
```

Generates code replacing `size_t` with `unsigned long`. Which is OK... until you try to build the generated sources in another platform (e.g. armhf, etc.) where `size_t` becomes something else, so the generated pattern code doesn't match anymore the original C++ API and fails to build. A solution might be re-running binder for each architecture, but I would like to avoid it.

## Questions

- Within binder, where is the type `size_t` and such converted into their underlying types?
- What would you think about a new config option to "freeze" certain types such that they are not resolved into their underlying types (if this is even possible...)

If you give me some pointers, I would be glad to attempt to implement it and PR.

CStewart-Burger:
I just came here for the same thing. I'm not particularly familiar with clang, but I think you'd want to look at  [https://github.com/RosettaCommons/binder/blob/67a9fe429698d752dc336122995f838bdeb7f92d/source/function.cpp#LL245C5-L245C5](url). 
I've also noticed some other argument types not being parsed as I'd expect. @jlblancoc , have you tested other stdint types?

jlblancoc:
Thanks for the pointer! I'll try to craft something around that function...

> , have you tested other stdint types?

sure, the same happen for examples for `uint8_t` which gets mapped into `unsigned char`, but that one is fairly standard and no problematic, but that's not always the case :-(

lyskov:
@jlblancoc , i just took a fresh look at the code and new API docs and i think there might be a way to get original type information! Let me play with this and i will get back to you.

jlblancoc:
Thanks, it would be great! In the meanwhile I tried with tricks redefining `__SIZE_TYPE__` and alike, but that's a dead end...

lyskov:
[...]

PR #259, opened 2023-05-16
Add new config option "+standard_type"

jlblancoc:
Follow up of #258

Feel free of changing the name "standard_type" to anything else...

### DIFF
--- a/source/type.cpp
+++ b/source/type.cpp
@@ -463,7 +463,29 @@ std::string standard_name(clang::QualType const &qt)
 
 
 	// if( begins_with(r, "std::") ) return r; //standard_name(r);
-	static std::set<string> standard_names {"std::size_t"};
+	static std::set<string> standard_names;
+
+	if (standard_names.empty())
+	{
+		standard_names =
+		{
+			// <cstddef>
+			"std::size_t", "std::ptrdiff_t",
+
+			// <stddef.h>
+			"size_t", "ptrdiff_t",
+
+			// <cstdint> (most common ones only)
+			"std::int8_t", "std::int16_t", "std::int32_t", "std::int64_t",
+			"std::uint8_t", "std::uint16_t", "std::uint32_t", "std::uint64_t",
+			"std::intmax_t", "std::uintmax_t", "std::intptr_t", "std::uintptr_t",
+
+			// <stdint.h> (most common ones only)
+			"int8_t", "int16_t", "int32_t", "int64_t",
+			"uint8_t", "uint16_t", "uint32_t", "uint64_t",
+			"intmax_t", "uintmax_t", "intptr_t", "uintptr_t",
+		};
+	}
 
 	if( standard_names.count(r) ) return r;
 	else return standard_name(qt.getCanonicalType().getAsString());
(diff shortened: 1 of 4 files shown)


source/type.cpp
```diff
@@ -461,14 +461,34 @@ std::string standard_name(clang::QualType const &qt)
...
+			"std::uint8_t", "std::uint16_t", "std::uint32_t", "std::uint64_t",
+			"std::intmax_t", "std::uintmax_t", "std::intptr_t", "std::uintptr_t",
+
+			// <stdint.h> (most common ones only)
+			"int8_t", "int16_t", "int32_t", "int64_t",
+			"uint8_t", "uint16_t", "uint32_t", "uint64_t",
+			"intmax_t", "uintmax_t", "intptr_t", "uintptr_t",
+		};
+		
+		for (const auto &t : Config::get().standard_types)
```
lyskov, line 489: pedantic and optional: would be good if project-standard placement of `const` is used, ie `auto const &`

source/type.cpp
```diff
@@ -461,14 +461,34 @@ std::string standard_name(clang::QualType const &qt)
...
+
+			// <cstdint> (most common ones only)
+			"std::int8_t", "std::int16_t", "std::int32_t", "std::int64_t",
+			"std::uint8_t", "std::uint16_t", "std::uint32_t", "std::uint64_t",
+			"std::intmax_t", "std::uintmax_t", "std::intptr_t", "std::uintptr_t",
+
+			// <stdint.h> (most common ones only)
+			"int8_t", "int16_t", "int32_t", "int64_t",
+			"uint8_t", "uint16_t", "uint32_t", "uint64_t",
+			"intmax_t", "uintmax_t", "intptr_t", "uintptr_t",
```
lyskov, line 486: Optional, if you feel up to this: would be great if small test using some (or maybe all?) of this types added into https://github.com/RosettaCommons/binder/tree/master/test . 🤔 I think something like function taking very long argument list containing all these types ~will do~ .

Plus class member function with similar list of arguments as well as public data members. That way we should cover all instances where type conversion is needed.

lyskov (commented):
Looks good to me, - thank you for taking care of this @jlblancoc ! Let me think for a bit if option name is best it could be. Otherwise this is good to be merged.

In light of your comment https://github.com/RosettaCommons/binder/issues/258#issuecomment-1550454963 - it might be a good idea to also add test case for, please see more comments on this below. Adding test is optional, - if you do not have bandwidth for this then i will add this later. Thanks,

jlblancoc:
Thanks for the quick fix @lyskov !

I tested it and indeed it seems to have worked well. I'm waiting for packages to be built in armhf to fully confirm it works now...

source/type.cpp
jlblancoc: Ok, I'll try it; if days pass by and have no feedback, feel free of merging and doing it yourself or stacking in the TO-DO ;-)

source/type.cpp
jlblancoc: sure, done now.

PR #265, opened 2023-09-06
Make CXX_STANDARD conditional on LLVM version

andrewandersonintel:
LLVM 16 requires C++17, since the LLVM core libraries now make use of lots of features available only from the C++17 standard and onwards.

If we successfully detect the LLVM_VERSION and it is >= 16, we can't successfully compile with --std=c++14, so set the standard version depending on LLVM_VERSION.


lyskov (approved):
Looks good to me, - thank you for taking care of this @andrewandersonintel !

Issue #266, opened 2023-10-01
build-and-run-tests.py: Missing positional argument in call to install_llvm_tool 


masinag:
Hi,
I get the following error when running `./build-and-run-tests.py`:
```Traceback (most recent call last):
  File "binder/./build-and-run-tests.py", line 53, in <module>
    main(sys.argv)
  File "binder/./build-and-run-tests.py", line 45, in main
    if not Options.binder: Options.binder = build.install_llvm_tool('binder', source_path+'/source', source_path + '/build', Options.binder_debug, jobs=Options.jobs, gcc_install_prefix=Options.gcc_install_prefix)
TypeError: install_llvm_tool() missing 1 required positional argument: 'compiler'
```

lyskov:
Thank you for letting us know and for the fix!

PR #267, opened 2023-10-01
Fix build scripts

masinag:
- fix missing parameter `compiler` in call to `build.install_llvm_tool` from `build-and-run-tests.py` (Issue #266)
- add parameter `compiler` to `build.get_cmake_compiler_options`. This  is needed since if the script is imported, then Options is not in global scope.


lyskov (approved):
LGTM! - thank you for fixing this @masinag !

Issue #263, opened 2023-08-31
Support for pybind11 smart_holder


kliegeois:
Hello,

For one of the projects for which I am using Binder, we need to use the smart_holder branch of pybindd11 ( https://github.com/pybind/pybind11/tree/smart_holder ). The differences, in term of pybindd11 code associated to the interface, are very minimal .

Would it be fine if I create during the next weeks a PR to binder to add an option to generate pybindd11 codes that use the smart_holder branch? Could this extra feature be accepted?

Thanks!

kliegeois:
Thanks @lyskov for your comments!

(This issue https://github.com/pybind/pybind11/pull/2839 is a good extra context of why it could be useful)

Up to now, I have been using the [conservative mode](https://github.com/pybind/pybind11/blob/smart_holder/README_smart_holder.rst#conservative-mode) of the smart holder branch.

So the code: 

```
#include <pybind11/pybind11.h>

struct Foo {};

PYBIND11_MODULE(example_bindings, m) {
    namespace py = pybind11;
    py::class_<Foo>(m, "Foo");
}
```

would now be replaced by:

```
#include <pybind11/smart_holder.h>

struct Foo {};

PYBIND11_SMART_HOLDER_TYPE_CASTERS(Foo)

PYBIND11_MODULE(example_bindings, m) {
    namespace py = pybind11;
    py::classh<Foo>(m, "Foo");
}
```

In my previous successful tests, I have been using `py::classh` and `PYBIND11_SMART_HOLDER_TYPE_CASTERS` only for the classes for which it was useful.

Concretely, for binder, I was thinking of adding a configuration option that can be enabled per class like this:
```
+use_smart_holder_for_class Foo
```
If the option is set, then we would add `PYBIND11_SMART_HOLDER_TYPE_CASTERS` and use `py::classh` for that class.
And if there is at least one `use_smart_holder_for_class`, we would use `#include <pybind11/smart_holder.h>` instead of `#include <pybind11/pybind11.h>`.

What do you think of this?

lyskov:
@kliegeois this looks quite reasonable to me, pull-request implementing this will be welcome! Could you please also include test in your PR? If so please take a look at https://github.com/RosettaCommons/binder/blob/master/test/T81.custom_trampoline_with_args.config test on how to specify specific config directive for particular test. Thanks,

Also, thank you for mentioning https://github.com/pybind/pybind11/pull/2839 issue! I noticed similar issue in my project and having general solution for them would great!

[...]

PR #269, opened 2023-10-17
Add support for pybind11  Smart_holder branch

kliegeois:
This PR adds support for the pybind11 smart_holder branch as discussed in #263. @lyskov


test/T61.smart_holder.ref
```diff
@@ -0,0 +1,126 @@
...
+#include <T61.smart_holder.hpp> // A
+#include <T61.smart_holder.hpp> // E1
+#include <T61.smart_holder.hpp> // E2_struct
+#include <T61.smart_holder.hpp> // E3_class
+#include <T61.smart_holder.hpp> // fi
+#include <T61.smart_holder.hpp> // fi_instantiated_by_use_in_function
+#include <sstream> // __str__
+
+#include <functional>
+#include <pybind11/smart_holder.h>
```
lyskov, line 11: Right now this test will fail since there is no `smart_holder.h` header in our Pybind11 fork... 🤔 - any thoughts how we can workaround this? Looking at https://github.com/pybind/pybind11/blob/smart_holder/include/pybind11/smart_holder.h it seems that it does include original `pybind11.h` but it is not clear why developers have decided to create separate header...

To get started, I see at least the following ways we can work around this issue:
[a] add smart_holder.h symlink pointing to original file to our current Pybind11 fork, - sounds like a error prone solution with some maintenance overhead 
[b] somehow mark this particular test as "special" so compilation failure is ignored
[c] allow user to specify which include to use in place of `smart_holder.h` and instruct test to override that to use standard header

@kliegeois thoughts on this? I am leaning toward [b] unless some there an elegant solution that I missing...

lyskov (commented):
Thank you for putting this together @kliegeois ! Looks good to me!

-- Could you please add minimal documentation for a new option? I think it would be great if it links to https://github.com/RosettaCommons/binder/issues/263 as well as to smart-holder Pybind11 branch. 
Thanks,

test/T61.smart_holder.ref
kliegeois: I would prefer the option [b], I think it is the cleanest option.

kliegeois:
Thanks @lyskov for the review!

I will add the documentation as requested.

source/config.hpp
```diff
@@ -52,6 +52,7 @@ class Config
 	string default_member_rvalue_reference_return_value_policy_ = "pybind11::return_value_policy::automatic";
 	string default_call_guard_ = "";
 	string holder_type_ = "std::shared_ptr";
+	string include_file_ = "pybind11/pybind11.h";
```
lyskov, line 55: Could you please rename this to something like `pybind11_include_file_`? Otherwise name will be quite confusing (ie include file for what?) since we talk about includes all over Binder code. Thanks,

source/config.hpp
```diff
@@ -90,6 +91,7 @@ class Config
 	string const &prefix_for_static_member_functions() { return prefix_for_static_member_functions_; }
 
 	string const &holder_type() const { return holder_type_; }
+	string const &include_file() const { return include_file_; }
```
lyskov, line 94: same as above: could you please rename this to `pybind11_include_file` to avoid confusion? Thanks,

test/T61.smart_holder.ref
lyskov: @kliegeois I gave this more thoughts and I think I have better idea: how about if we add one more config directive: `pybind11_include_file` to allow user to customize it and in the same time remove automatic changing `pybind11_include_file` value when one or more `smart_holder` config options was specified? - I think that will give very nice and predictable behavior as well as expand possible code use (for example if for some reason someone need to use different pybind11 header for other reasons). And in the same time this will allow us to fully run the test. Thoughts?

lyskov (changes requested):

kliegeois:
Thanks for the review @lyskov !
I updated the PR as suggested.

lyskov (approved):
LGTM!

lyskov:
@kliegeois thank you for putting this together! I really like how this is implanted now, - will merge this shortly!

PR #272, opened 2023-11-15
update build instructions for macos

nikitalita:
Needed to add arm64 instructions.
Added the least recent version of LLVM for arm64 macos (14.0.6) that I could find because I wasn't sure which LLVM versions that this supported; the binder executable builds fine, at least.


lyskov (approved):
LGTM, - thank you @nikitalita for adding this!

PR #274, opened 2023-11-15
Fix macos tests

nikitalita:
macos tests have been failing for quite a while now but weren't being detected.

pybind 11 2.10 removed python 2 support, and since pybind11 got updated to >= 2.10 in homebrew, all the python 2 tests have been failing. The script didn't actually exit with an error code if the tests failed though, so it wasn't being detected.

Additonally, tests are failing to build at all on macos 13 due to differences in the header file layouts; if we don't include the C header files AFTER the C++ header files, binder will fail to find c std header files (e.g. `<cstdargs>`->`stdargs.h`). I have modified tests/CMakeLists.txt to fix this.


lyskov (approved):
Looks good to me, - thank you for catching this @nikitalita !

PR #275, opened 2023-11-15
Fix Python finding 

nikitalita:
The current `FindPython3.cmake` had a bug in which it failed to find simplified python exes, like `python3`.

Additionally, it was not very good at dealing with systems that had numerous python installations to choose from, like my own; `cmake` >= 3.19 has `FindPython3`, and all the other cmake Modules that you currently have. So, I changed Cmakelists.txt to only include the modules if the cmake version is < 3.19.


lyskov (approved):
LGTM, - thank you @nikitalita !

PR #277, opened 2023-12-18
Fix external operators in limitations documentation

haiiliin:
The supported [external operators](https://cppbinder.readthedocs.io/en/latest/limitations.html#external-operators) are ignored in the documentation because a blank line is missing the `code-block` directive.
image


lyskov (approved):
LGTM, - thank you for catching this @haiiliin !

PR #281, opened 2024-02-01
Implement skip-line-number option

andriish:
Implementation of an option to skip line numbers in the comments to the generated code.

This option is useful for the following use case:
 - the bindings are stored in e.g. git repository and are regenerated from time to time

In this scenario, the input headers might be updated, e.g. the new functions can be added without the updates of the signatures of the binded functions. As a result, if one reruns the binder, the only update of the binded code might be the changes in the line numbers.

To avoid this situation one can remove the line numbers from the comments to the code.

### DIFF
--- a/documentation/config.rst
+++ b/documentation/config.rst
@@ -33,9 +33,9 @@ files. Typically the following files will be generated: ``<root-module>.cpp``, `
 
 ``--flat`` if specified instruct Binder to write generate code files into single directory. Generated files will be named as ``<root-module>.cpp``, ``<root-module>_1.cpp``, ``<root-module>_2.cpp``, ... etc.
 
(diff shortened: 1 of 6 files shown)


lyskov (approved):
- i like this idea @andriish ! Please let me know when you are ready to merge this (looks good to me, assuming build tests passed). I think i will be using it in self-test.py script right away.

andriish:
Hi @lyskov, great! Yes, the PR is ready to go.

lyskov (dismissed):
- Looks good to me, - thank you for adding this @andriish ! I can merge this as-is. I wonder if you mind adding one-line documentation into https://github.com/RosettaCommons/binder/blob/master/documentation/config.rst ? 
Thanks,

lyskov (approved):
LGTM, - thank you for adding this @andriish !

PR #282, opened 2024-02-13
Update util.cpp to avoid compiler warning and  trow an error if a directory cannot be created.

andriish:
Update util.cpp to avoid compiler warning 
```
/source/util.cpp:138:15: warning: ignoring return value of 'int system(const char*)' declared with attribute 'warn_unused_result' [-Wunused-result]
  138 |         system(command_line.c_str());
```

and  trow an error if a directory cannot be created.


lyskov (approved):
LGTM! - thank you @andriish !

PR #283, opened 2024-02-13
Update class.cpp to avoid compiler warnings

andriish:
Update class.cpp to avoid compiler warnings

```
source/class.cpp:667:1: warning: multi-line comment [-Wcomment]
  667 | // #define PYBIND11_OVERLOAD_INT(ret_type, cname, name, ...) { \
      | ^
```

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -662,27 +662,6 @@ bool is_callback_structure_constructible(CXXRecordDecl const *C)
 	return true;
 }
 
-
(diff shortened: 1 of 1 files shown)


lyskov (changes requested):
Hm.. I want to keep this for reference, - do you think we can around the warning by using range comment `/*...*/`?

andriish:
See #285

PR #286, opened 2024-03-13
Modernization of CI

andriish:
Modernization of CI; use matrix instead of separate Docker files.


lyskov (approved):
Looks good to me, - thank you for taking care of this @andriish !

PR #291, opened 2024-04-04
Update class.cpp to avoid else after return

andriish:
Update class.cpp to avoid else after return


andriish:

andriish:
The rule  clang complaining about is here: 

https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

This in in the *generated* code.

andriish:
@lyskov

lyskov (approved):
LGTM, - thank you @andriish !

PR #292, opened 2024-04-04
Update context.cpp to avoid else after return

andriish:
Update context.cpp to avoid else after return


andriish:

lyskov (commented):
Hm... i do like the explicit `else` in https://github.com/RosettaCommons/binder/pull/292/files#diff-a991df2f19ad54fba2d1e4aa256a303fb097f489381a1adf8c5319726132c651L113 and i believe other changes now merged?

andriish:
The rule is here:

https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

But for the automatically generated code it also makes sense to reduce the size.

lyskov:
Oh, i have not realized that this is inside generated code, - i see your point, - makes sense to me!

lyskov (approved):
LGTM, - thank you @andriish !

PR #294, opened 2024-04-04
Update context.cpp use empty()

andriish:
Update context.cpp use empty() to avoid clang-tidy warnings


andriish:

source/context.cpp
```diff
@@ -116,7 +116,7 @@ PYBIND11_MODULE({2}, root_module) {{
 
 	std::vector< std::pair<std::string, std::string> > sub_modules {{
 {3}	}};
-	for(auto &p : sub_modules ) modules[p.first.size() ? p.first+"::"+p.second : p.second] = modules[p.first].def_submodule( mangle_namespace_name(p.second).c_str(), ("Bindings for " + p.first + "::" + p.second + " namespace").c_str() );
+	for(auto &p : sub_modules ) modules[ !p.first.empty() ? p.first+"::"+p.second : p.second] = modules[p.first].def_submodule( mangle_namespace_name(p.second).c_str(), ("Bindings for " + p.first + "::" + p.second + " namespace").c_str() );
```
lyskov, line 119: Out of curiosity: what warning give clang-tidy? The `string::size` have constance complexity since C++11 (https://en.cppreference.com/w/cpp/string/basic_string/size) so it should be fine 🤔

andriish: This one:
https://clang.llvm.org/extra/clang-tidy/checks/readability/container-size-empty.html

Makes sense and does not require conversion.

lyskov: Ah, i see, so i does not "know" that this is a string oh well, - i generally do like to use `empty` if possible so yeah, lets refactor this. However it will best to swap then/else blocks and remove negation to improve readability: 
`!p.first.empty() ? p.first+"::"+p.second : p.second` → 
`p.first.empty() ? p.second : p.first+"::"+p.second`, - would you mind doing that? Thanks,

andriish: Yes.

andriish:
@lyskov

lyskov (approved):
LGTM, - thank you @andriish !

PR #296, opened 2024-04-06
LLVM 18 compatibility

andriish:
LLVM 18 compatibility

PR #297, opened 2024-04-06
Update context.cpp to comply with pybind11 2.0+ docs

andriish:
Update context.cpp to comply with pybind11 2.0+ docs, i.e.
PYBIND11_DECLARE_HOLDER_TYPE from version 2.0 requires at least 3 arguments otherwise the compillers will issue warnings.
See https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html
```
The first argument of PYBIND11_DECLARE_HOLDER_TYPE() should be a placeholder name that is used as a template parameter of the second argument. Thus, feel free to use any identifier, but use it consistently on both sides; also, don’t use the name of a type that already exists in your codebase.

The macro also accepts a third optional boolean parameter that is set to false by default. 
```


andriish:

lyskov (approved):
LGTM, - thank you @andriish !

Issue #298, opened 2024-04-09
Wrong code for MSVC


andriish:
In some cases binder will generate code that is not compatible with MSVC, e.g.

```
using _binder_ret_0 = class std::map<std::string, std::string >;
```

The following  code is will not compile in MSVC:

```
#include <stdexcept>
#include <string>
#include <map>
using _binder_ret_0 = class std::map<std::string, std::string >;
int main()
{
    _binder_ret_0 a;
    a["a"]="a";
    return 0;
}
```
but is OK for GCC.

To make the code compatible one should remove the "class" word.

```
using _binder_ret_0 = std::map<std::string, std::string >;
```

lyskov:
resolved in https://github.com/RosettaCommons/binder/pull/299

PR #299, opened 2024-04-09
Update class.cpp to make code compatible with MSVC

andriish:
Update class.cpp to make code compatible with MSVC.

Should solve #298

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -752,7 +752,7 @@ string bind_member_functions_for_call_back(CXXRecordDecl const *C, string const 
 				if( return_type.find(',') != std::string::npos ) {
 					string return_type_alias = "_binder_ret_" + std::to_string(ret_id);
 					++ret_id;
-					if (return_type.rfind("class ", 0) == 0) return_type = return_type.substr(6);
+					if (begins_with(return_type,"class ")) return_type = return_type.substr(6);
 					c += "\tusing {} = {};\n"_format(return_type_alias, return_type);
 					return_type = std::move(return_type_alias);
 				}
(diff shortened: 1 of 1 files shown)


lyskov (changes requested):
Hm... @andriish i do not think this code is doing what you intended (see https://en.cppreference.com/w/cpp/string/basic_string/rfind). Please instead call  `begins_with` from https://github.com/RosettaCommons/binder/blob/master/source/util.cpp#L88C6-L88C17
Thanks,

andriish:
Hi @lyskov,
the code does what is expected:
```
$ root -l
root [0] std::string return_type="class XXX"
(std::string &) "class XXX"
root [1] if (return_type.rfind("class ", 0) == 0) return_type = return_type.substr(6);
root [2] return_type
(std::string &) "XXX"
root [3] 
```
I havent seen your `begins_with`, will use it.
But I would actually update `begins_with`.

Andrii

andriish:
@lyskov

lyskov:
> ```
> $ root -l
> root [0] std::string return_type="class XXX"
> (std::string &) "class XXX"
> root [1] if (return_type.rfind("class ", 0) == 0) return_type = return_type.substr(6);
> root [2] return_type
> (std::string &) "XXX"
> root [3] 
> ```
-- oh, i see, so the `pos` argument is offset from the _beginning of the string_ and not the end 🤦🏻‍♂️, - yeah, that will work too!
Btw, do you use some kind of interpreter to test this, which one? 

> I havent seen your `begins_with`, will use it. But I would actually update `begins_with`.
-- could you please elaborate? Ie do you expect `rfind` implementation to be more efficient or something? 
Thanks,

lyskov (approved):
LGTM, - thank you @andriish !

lyskov:
> I havent seen your begins_with, will use it. But I would actually update begins_with.

@andriish could it be that you are referring to the old study that claims that for natural languages it yield better performance to always search from the end of the strings and not beginning (ie there is less number of strings that ends the same compare to number of strings that have begin with same sequence)?

andriish:
-- The interpreter is https://root.cern/
 -- The  "rfind" is  just "oneliner for string::starts_with" implemented in C++11

PR #303, opened 2024-06-19
Setup binder as a Python package and publish pre-built wheels to PyPI

haiiliin:
## Description of the implementation

- The python package is placed in the `python` directory to avoid conflicts.
- The `llvm-project` branch `v6.x` is submoduled in `python/llvm-project` and is included in `python/CMakeLists.txt` with `add_subdirectory`.
- The `source` directory is symbolically linked to `python/llvm-project/clang-tools-extra/binder` in the [llvm-project](https://github.com/haiiliin/llvm-project) submodule.
- The version of the python package is automatically inferred from the git tag using `setuptools-scm`.
- `scikit-build-core` is used to build the Python package with cmake.
- `cibuildwheel` is used to build wheels on different platforms, and archs using GitHub actions, and they will be published to PyPI when a release is published.


haiiliin:
Linux build failed because it is built in a docker container. Part of (the python directory) the project was copied in the container and the symbolic link for the binder source broke.

haiiliin:
Looks good now

haiiliin:
Notes on publishing to PyPI: please give me a PyPI account name so I can invite you as a collaborator for the [cppbinder](https://pypi.org/project/cppbinder/) project. You can create an API token in PyPI and add it as a secret `PYPI_API_TOKEN` in the repository settings.

haiiliin:
@lyskov Can you take a look at this?

documentation/install.rst
```diff
@@ -1,5 +1,18 @@
-Installation
-============
+Installation with pre-built wheels
+==================================
+
+The easiest way to install Binder is to use pre-built wheels from PyPI. To install Binder, run:
+
+.. code-block:: bash
+
+    pip install cppbinder
```
lyskov, line 8: Have you checked how well this works when deployment system is different (ie lib stdc++ is different) from the the one you build wheels one? For my projects this was significant issue so i end up just rebuilding Binder on target system.

lyskov:
@haiiliin this is interesting idea, let me think about this!

Also, i see that you added LLVM as submodule. Two thought on this: 
* it probably best to use official LLVM repos from https://github.com/llvm/llvm-project
* It might be best move all of this to a separate repository that have both Binder and and LLVM repos as submodules. That way it will be clear separation of responsibility. Thoughts?

Could you please let me know if you tested pre-build wheels on Linux distros with different libstdc++ versions? (different from the one you build the package). In the past this was an issue so i am suspecting it still is. 

Thanks,

lyskov:
@haiiliin to clarify:
> Could you please let me know if you tested pre-build wheels on Linux distros with different libstdc++ versions? (different from the one you build the package). In the past this was an issue so i am suspecting it still is. 

-- particularly that the code generated by Binder (say when running self test involving C++ std lib) is indeed compilable on the target systems. 

Thanks,

haiiliin:
> * it probably best to use official LLVM repos from https://github.com/llvm/llvm-project

This reason for not using the official LLVM project is that we have to create a symbolic link to the binder source inside the LLVM project and then add binder as a sub project in LLVM. I using the official LLVM project, when someone clone the binder project, the command to install the package locally `pip install ./python` will fail. But there maybe a solution for this that we can run some scripts to create the symbolic links before build the package in `scikit-build-core`. But it requires administrator access to install the package locally on windows.

> * It might be best to move all of this to a separate repository that have both Binder and and LLVM repos as submodules. That way it will be clear separation of responsibility. Thoughts?

The problem is that when a release is published in this project, we have to update submodules in another repository and then publish a same release again which makes it quite annoying.

> Could you please let me know if you tested pre-build wheels on Linux distros with different libstdc++ versions? (different from the one you build the package). In the past this was an issue so i am suspecting it still is.

I think [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/) (which uses the use the [manylinux](https://github.com/pypa/manylinux) docker images to build wheels on different archs and glibc versions) will take care of this to make the wheels to support most of linux distributions. 

I only used the wheel built on `ubuntu-22.04` using GitHub actions and tested on WSL `ubuntu-24.04`.

haiiliin:
By default, cibuildwheel only builds `x86_64` and `i686` wheels on linux, if you think wheels need to be built, we can also build `aarch64`, `ppc64le`, `s390x` wheels

See here: https://cibuildwheel.pypa.io/en/stable/options/#archs

lyskov:
Great, - i see that you already pushed some of the binaries to PyPI which makes things easier to test. I have tried running Binder from these binaries on various Linux distos and so far things working!

One issue that i noticed is that it looks like right now wheel packages contain only the Binder executable and no include files. This create rather annoying usability issues since user will have to manually setup all include path for system (ie C++ std lib) include. Have you consider packing includes alongside with Binder Binaries? Steps to reproduce if you curious:
1. run ubuntu:20.04 docker image
2. install python3, python3-pip packages
3. pip install cpp binder
4. clone binder and pybind11 into /root
5. cd /root/binder/test && ./self-test.py --binder=`which binder` --pybind11 /root/pybind11/include/ --accept


re submodules: i have give this issue some serious consideration and i see a few tech issues for current setup:
> The problem is that when a release is published in this project, we have to update submodules in another repository and then publish a same release again which makes it quite annoying.

-- this is indeed and inconvenience but i think there is an easy way to fix it. How about if this new project will not have a submodules but rather do a fresh clones of both LLVM and Binder before builds? One particular issue with current setup is that adding LLVM as submodule is problematic (due to bulk size that it adds) for all projects who do not really need this functionality. This will became particularly an issue for projects who already have Binder as submodules (along other submodules) (my main project is setup this way and i know at least few more who have similar setup so this is a real use cases) which will require them to deal with recursive submodules initialization and how to workaround it.

-- another issue is chain of responsibility, right now Binder repo is clearly about Binder and not various way we can package it. This feels like right approach to me (ie keep things separated when possible). Having PyPI related things inside Binder repo while convenient at some points also merge maintenance of Binder and Binder-wheels-in-PyPI together which i do not think is a good move in the long run. For instance if we have this project separated then person who maintain wheels does not need to be the same person who maintain Binder (for example it could be you).

I guess in my mind PyPI packages is distribution system similar how one might create package for Debian, MacPorts or Spack. So lets keep this things separated unless there is a good tech point why they should not.

Thanks,

source/CMakeLists.txt
```diff
@@ -122,7 +122,11 @@ else()
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${LLVM_COMPILE_FLAGS}")
+  list(REMOVE_ITEM CMAKE_CXX_FLAGS "-fno-exceptions")
+  string (REPLACE "-fno-exceptions" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+  message(STATUS "binder: CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}")
 endif()
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/binder
+install(TARGETS binder
 PERMISSIONS WORLD_EXECUTE WORLD_READ OWNER_WRITE OWNER_READ OWNER_EXECUTE
-DESTINATION ${CMAKE_INSTALL_BINDIR})
+DESTINATION binder/bin COMPONENT binder)
```
andriish, line 132: Please leave the ${CMAKE_INSTALL_BINDIR} as it is. The 'COMPONENT' is ok though.

python/tests/test_binder.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def test_check_binder_llvm_version():
+    # Check the output of the `binder -version` command to obtain the versions of Binder and LLVM
+    output = subprocess.check_output(["binder", "-version"]).decode()
+    pattern = r"binder ([\d\.]+)\nLLVM ([\d\.]+)"
+    binder_version, llvm_version = re.match(pattern, output).groups()
+
+    # Check the versions of LLVM
+    assert llvm_version == "6.0.1", f"LLVM version mismatch: {llvm_version} != 6.0.1"
```
andriish, line 12: Why '6.0.1' is special?

python/pyproject.toml
```diff
@@ -0,0 +1,96 @@
...
+wheel.expand-macos-universal-tags = true
+
+[[tool.scikit-build.generate]]
+path = "binder/_version.py"
+template = '''
+version = "${version}"
+'''
+
+[tool.cibuildwheel]
+build = "cp312-*"
```
andriish, line 75: 312 should not be hardcoded.

haiiliin: Since the package does not depend on the Python version, we can only use a specific Python version to build the wheel, the output wheel will support any other Python versions.

python/tests/test_binder.py
haiiliin: It is the LLVM version submoduled in the project.

haiiliin:
The LLVM project is now cloned at build time

Issue #85, opened 2019-10-21
Making Binder aware of Eigen


cuihantao:
Hi there,

First of all, thanks for putting together this great tool, Binder. After a quick configuration, I was able to generate the binding code that would take days to write by hand. 

## Problem 
One issue I have with the generated code is the awareness of Eigen. Pybind11 has supported Eigen pretty well for the daily work (See https://pybind11.readthedocs.io/en/stable/advanced/cast/eigen.html). 

Binder doesn't seem to recognize Eigen (or its namespace). If forced to do `-namespace Eigen`, it will omit all functions that has an Eigen argument. Otherwise, it will place Eigen files under `unknown` and produce a ton of warning, which looks like

```
Warning: file_name_prefix_for_binder could not determent file name for decl: Eigen::internal::Packet2cf, result is too short!
```
The compilation fails afterward.

## Reproduce
The attached is a minimum version of code using Eigen in a struct, modified from the `test_struct` example. A CMake build script is utilized. Note that the code will automatically clone Eigen into the folder root directory. You will also need to change a few hard-coded binder paths at the beginning of the script.

Kindly let me know if there's any workaround as well.

## Attachment
[binder.zip](https://github.com/RosettaCommons/binder/files/3751933/binder.zip)

lyskov:
@cuihantao the warning above is just that - a warning, it should be ignored. It is emitted only to indicated that some of the type/functions could not be traced to it .hpp files. This usually happens when C++ uses relative includes like `#include "aa.hh"`. In such cases Binder can not determent location of include files and so compilation will fail (this limitation is documented here: https://cppbinder.readthedocs.io/en/latest/basics.html). You can workaround this issue by using `+include` directive in Binder config and force inclusion of some of Eigen headers.

re awareness of Eigen: - not sure what exactly do you mean by this but in general no special handling of third-party C++ libs should be needed: Binder will treat any library calls the same as any other C++ call. If you want to bind Eigen types in Python then you need to add `+namespace Eigen` directives into Binder config. This will trigger bindings generation for all functions/types in Eigen namespace.

cuihantao:
Thanks for the reply. I guess I will try to include all the header files from Eigen. I will post an update once done.

shiveshkhaitan:
@cuihantao were you able to figure out how to use Eigen with binder? I am facing exactly the same issue that you have mentioned.

cuihantao:
[...]

PR #304, opened 2024-06-25
Do not check type dependencies when determine if bindings should be skipped

haiiliin:
## Description

I am using binder to generate bindings for functions/methods that take Eigen matrices as arguments or return values. For example:

```c++
#include <Eigen/Dense>

void eigen_argument_function(const Eigen::Matrix<T, 3, 3> &m) {};
```

I don't want to bind the whole Eigen library, thus I have to exclude the `Eigen` namespace for binder:
```
- namespace Eigen
```

However, the bindings of functions/methods that take Eigen matrices as arguments or return values do not generate because these functions/methods depend on the eigen library and thus skipped. Bindings of any functions/methods that have dependencies of excluded namespaces will not be generated, and this is not what I wanted.

In my own understanding, when I exclude a namespace, I mean I only want to exclude classes/functions in that namespace, but not classes/functions that have dependencies on that namespace.

In addition, I found that when generating bindings for enums, the dependencies will not be checked:
https://github.com/RosettaCommons/binder/blob/7d6ec91db99d5f049db7dd8aea186a4c74f6742f/source/enum.cpp#L59-L67
https://github.com/RosettaCommons/binder/blob/7d6ec91db99d5f049db7dd8aea186a4c74f6742f/source/enum.cpp#L69-L82

If this change is not appropriate, maybe adding a configuration option would be helpful.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -340,7 +340,8 @@ bool is_binding_requested(clang::CXXRecordDecl const *C, Config const &config)
 	if( (not bind_class_template_specialization) and dyn_cast<ClassTemplateSpecializationDecl>(C) ) return false;
 
 	bool bind = config.is_namespace_binding_requested(namespace_from_named_decl(C));
-	for( auto &t : get_type_dependencies(C) ) bind &= !is_skipping_requested(t, config);
+	// Do not check type dependencies when checking if binding is requested
+	// for( auto &t : get_type_dependencies(C) ) bind &= !is_skipping_requested(t, config);
 	return bind;
 }
 
@@ -358,7 +359,8 @@ bool is_skipping_requested(clang::CXXRecordDecl const *C, Config const &config)
 
 	bool skip = config.is_namespace_skipping_requested(namespace_from_named_decl(C));
 
-	for( auto &t : get_type_dependencies(C) ) skip |= is_skipping_requested(t, config);
+	// Do not check type dependencies when checking if binding is requested
+	// for( auto &t : get_type_dependencies(C) ) skip |= is_skipping_requested(t, config);
 
 	return skip;
 }
--- a/source/function.cpp
+++ b/source/function.cpp
@@ -318,7 +318,8 @@ bool is_binding_requested(FunctionDecl const *F, Config const &config)
 
 	bool bind = config.is_namespace_binding_requested(namespace_from_named_decl(F));
 
-	for( auto &t : get_type_dependencies(F) ) bind |= binder::is_binding_requested(t, config);
+	// Do not check type dependencies when checking if binding is requested
+	// for( auto &t : get_type_dependencies(F) ) bind |= binder::is_binding_requested(t, config);
 
 	return bind;
 }
@@ -353,7 +354,8 @@ bool is_skipping_requested(FunctionDecl const *F, Config const &config)
 	}
 	// outs() << "OK\n";
 
-	for( auto &t : get_type_dependencies(F) ) skip |= is_skipping_requested(t, config);
+	// Do not check type dependencies when checking if binding is requested
+	// for( auto &t : get_type_dependencies(F) ) skip |= is_skipping_requested(t, config);
 
 	return skip;
 }


lyskov:
@haiiliin i see your point. Current design and behavior is intentional though. Here is how it ""suppose to work": explicitly trigger bindings of namespaces that you want to bind and do not mention (either with `+namespace` or `-namespace`) namespaces that you do not mind being bound but for which you do not bindings by default, ie like the bindings for Eigen library in your case. 

So say your main project is reside in C++ namespace `aaaa`, then we specify `+namespace aaaa` and do not mention `eigen` namespace at all. The result should be that only function/classes from Eigen that got used in your main project will be bound. Have you tried this approach?

If you curios: reason for current design is that `-` namespace/class/function options is viewed as "direct" orders to skip particular objects no matter what.

haiiliin:
> So say your main project is reside in C++ namespace `aaaa`, then we specify `+namespace aaaa` and do not mention `eigen` namespace at all. The result should be that only function/classes from Eigen that got used in your main project will be bound. Have you tried this approach?

I tried this approach, but it turned out the bindings of the eigen library are still generated in the `unknown` directory, same as the situation in #85.

lyskov:
Do you see _some_ binding generated for Eigen or do you have reason to believe that Binder is trying to bind all the Eigen code? The former is expected behavior while later could be indication of bug in Binder. Also, could you please cut-paste `+/-namespace` directives from you Binder config here?

Re placement bindings in `unknown` dir: i am guessing the Eigen library using comma-based incudes which usually make Binder fall back to auto generating name source files in unknown dir. This by itself is not a sign of any problem.

haiiliin:
Binder tries to generate some bindings for the library. For simple eigen library usage, the binder only generates several unknown files. However, my project generates almost 5000 unknown files. I don't use many of the eigen library, just the eigen matrix, some of its methods, and the linear algebra functions.

Neither way, the build will fail. https://github.com/haiiliin/binder-eigen-test is a project for the testing, please check it out. 

Here are some logs of the build:

```
[1/2] Building CXX object bindings\CMakeFiles\bindings.dir\unknown\unknown.cpp.obj
FAILED: bindings/CMakeFiles/bindings.dir/unknown/unknown.cpp.obj 
C:\PROGRA~1\MIB055~1\2022\COMMUN~1\VC\Tools\MSVC\1437~1.328\bin\Hostx64\x64\cl.exe  /nologo /TP -Dbindings_EXPORTS -IC:\Users\hailin\Desktop\binder-eigen-test\deps\eigen -IC:\Users\hailin\Desktop\binder-eigen-test\include -external:IC:\Users\hailin\AppData\Local\Programs\Python\Python312\include -external:IC:\Users\hailin\AppData\Local\Programs\Python\Python311\include -external:W0 /DWIN32 /D_WINDOWS /EHsc /Zi /Ob0 /Od /RTC1 -MDd /bigobj /showIncludes /Fobindings\CMakeFiles\bindings.dir\unknown\unknown.cpp.obj /Fdbindings\CMakeFiles\bindings.dir\ /FS -c C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\cast.h(746): error C2338: static_assert failed: 'Holder classes are only supported for custom types'
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\cast.h(820): note: see reference to class template instantiation 'pybind11::detail::copyable_holder_caster<base,holder,void>' being compiled
        with
        [
            base=Eigen::Matrix<float,3,3,0,3,3>,
            holder=std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>
        ]
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\type_traits(1312): note: see reference to class template instantiation 'pybind11::detail::type_caster<holder,void>' being compiled
        with
        [
            holder=std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>
        ]
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\cast.h(867): note: see reference to class template instantiation 'std::is_base_of<pybind11::detail::copyable_holder_caster<base,holder,void>,pybind11::detail::type_caster<holder,void>>' being compiled
        with
        [
            base=Eigen::Matrix<float,3,3,0,3,3>,
            holder=std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>
        ]
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\detail/common.h(863): note: see reference to class template instantiation 'pybind11::detail::is_holder_type<type_,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>' being compiled
        with
        [
            type_=Eigen::Matrix<float,3,3,0,3,3>
        ]
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\pybind11.h(1511): note: see reference to class template instantiation 'pybind11::detail::exactly_one<pybind11::class_<Eigen::Matrix<float,3,3,0,3,3>,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>::is_holder,std::unique_ptr<Eigen::Matrix<float,3,3,0,3,3>,std::default_delete<Eigen::Matrix<float,3,3,0,3,3>>>,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>' being compiled
C:\Users\hailin\AppData\Local\Programs\Python\Python311\include\pybind11\pybind11.h(1511): note: see reference to alias template instantiation 'pybind11::detail::exactly_one_t<pybind11::class_<Eigen::Matrix<float,3,3,0,3,3>,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>::is_holder,std::unique_ptr<Eigen::Matrix<float,3,3,0,3,3>,std::default_delete<Eigen::Matrix<float,3,3,0,3,3>>>,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>' being compiled
C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp(18): note: see reference to class template instantiation 'pybind11::class_<Eigen::Matrix<float,3,3,0,3,3>,std::shared_ptr<Eigen::Matrix<float,3,3,0,3,3>>>' being compiled
C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp(43): error C2440: 'type cast': cannot convert from 'overloaded-function' to 'Eigen::Matrix<float,3,3,0,3,3> &(__cdecl Eigen::Matrix<float,3,3,0,3,3>::* )(const Eigen::Matrix<float,3,3,0,3,3> &) const'
C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp(43): note: None of the functions with this name in scope match the target type
C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp(84): error C2440: 'type cast': cannot convert from 'overloaded-function' to 'Eigen::Matrix<double,3,3,0,3,3> &(__cdecl Eigen::Matrix<double,3,3,0,3,3>::* )(const Eigen::Matrix<double,3,3,0,3,3> &) const'
C:\Users\hailin\Desktop\binder-eigen-test\bindings\unknown\unknown.cpp(84): note: None of the functions with this name in scope match the target type
ninja: build stopped: subcommand failed.
```

lyskov:
re 5k files: this most likely is due to dependencies chain: ie your project used the some of the Matrix classes which in turn uses some other classes and so on. If this is absolutely not what you want then perhaps using your proposed path for your projects is the best approach. I certainly have similar cases in the past when for some projects it was not unclear how bind them without introducing some project specific features into Binder 🤷🏻‍♂️.

Re errors: which Pybind11 version do you use? If this is upstream then i would recommend to try use one from our fork and see if that makes a difference. Also, it might worth to check if generated code look correct and if yes then propagate this to Pybind11 team.

Hope this helps,

PR #308, opened 2024-12-17
Remove import of removed, unused Python modules

adamnovak:
Python 3.12 doesn't have `imp`, and this file doesn't seem to actually use it.

### DIFF
--- a/build.py
+++ b/build.py
@@ -14,7 +14,7 @@
 
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


adamnovak:
It looks like some exciting but unrelated things have gone wrong with the CI on this, which look a lot like what went wrong lin the latest build of the main branch.

lyskov (approved):
Looks good to me, - thank you @adamnovak !

PR #309, opened 2024-12-18
Support Ubuntu 24.04 and LLVM 14.0.5; automatically select LLVM version based on compiler version

adamnovak:
This PR updates Binder to work on Ubuntu 24.04, with its GCC 13.

It also implements the compiler-version-dependent LLVM version selection logic that the build scripts wanted, and adds an `--llvm-version` command line flag to them to override it.

It includes LLVM 14.0.5, which will be automatically selected for GCC 13 or newer, since it's the oldest LLVM I could find that can build on GCC 13 and understand GCC 13's STL headers.

I had to change how the LLVM tarballs are extracted to not lose the `cmake` directory that LLVM 14.0.5 needs next to its source directory to build. I added some version tracking for this directory, which will involve redownloading LLVM if it needs to change versions.

### DIFF
--- a/build-and-run-tests.py
+++ b/build-and-run-tests.py
@@ -33,6 +33,7 @@ def main(args):
     parser.add_argument('--binder', default='', help='Path to Binder tool. If none is given then download, build and install binder into main/source/build. Use "--binder-debug" to control which mode of binder (debug/release) is used.')
     parser.add_argument("--binder-debug", action="store_true", help="Run binder tool in debug mode (only relevant if no '--binder' option was specified)")
     parser.add_argument('--pybind11', default='', help='Path to pybind11 source tree')
+    parser.add_argument('--llvm-version', default=None, choices=['6.0.1', '13.0.0'], help='Manually specify LLVM version to install')
     parser.add_argument('--annotate-includes', action="store_true", help='Annotate includes in generated source files')
     parser.add_argument('--trace', action="store_true", help='Binder will add trace output to to generated source files')
     parser.add_argument('--gcc-install-prefix', default=None, help='Path to GCC install prefix which will be used to determent location of libstdc++ for Binder build. Default is: auto-detected. Use this option if you would like to build Binder with compiler that was side-installed and which LLVM build system failed to identify. To see what path Binder uses for libstdc++ run `binder -- -xc++ -E -v`.')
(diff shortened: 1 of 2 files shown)


build-and-run-tests.py
```diff
@@ -33,6 +33,7 @@ def main(args):
     parser.add_argument('--binder', default='', help='Path to Binder tool. If none is given then download, build and install binder into main/source/build. Use "--binder-debug" to control which mode of binder (debug/release) is used.')
     parser.add_argument("--binder-debug", action="store_true", help="Run binder tool in debug mode (only relevant if no '--binder' option was specified)")
     parser.add_argument('--pybind11', default='', help='Path to pybind11 source tree')
+    parser.add_argument('--llvm-version', default=None, choices=['6.0.1', '13.0.0', '14.0.5'], help='Manually specify LLVM version to install')
```
lyskov, line 36: Do you think we can avoid using choices so user can install arbitrary LLVM version? Note that i am not saying that we should do that, just raising the possibility since this approach will be most flexible.

adamnovak: Probably, but the build system changes between major versions, so we'd need to have logic to parse the version and figure out what operations we think are going to be required to build that version.

lyskov: Fair point! - i guess let's merge this as-is then. Let re-read changes one more time and i will get back to you.

build.py
```diff
@@ -129,15 +163,26 @@ def install_llvm_tool(name, source_location, prefix_root, debug, compiler, jobs,
...
             '13.0.0' : ('https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/llvm-13.0.0.src.tar.xz', 'https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang-13.0.0.src.tar.xz'),
+            '14.0.5' : ('https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.5/llvm-14.0.5.src.tar.xz', 'https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.5/clang-14.0.5.src.tar.xz')
         }[llvm_version]
 
-        if not os.path.isfile(prefix + '/CMakeLists.txt'):
+        # The LLVM tarballs may include a "cmake" directory that needs to be
+        # sibling to the main source directories in order for the build to
+        # work. That path can't depend on the LLVM version, so we track that ourselves.
+        cmake_path = os.path.join(prefix_root, "cmake")
+        cmake_version_path = os.path.join(cmake_path, "llvm_version.txt")
```
lyskov, line 173: Do we really need this new "marker" (ie llvm_version.txt file)? We already have llvm_version encoded in our signature file, - is not that enough? (see https://github.com/RosettaCommons/binder/pull/309/files#diff-b70bfb027a8ef3ec4ab07a53a3b8ba83e63334ba3ce5ca6d9d773900ed452e5eL113). Please elaborate, Thanks,

lyskov (commented):
Looks good to me with few minor questions (below).

Also, have you tested this locally? If so on which platforms? thanks,

adamnovak:
@lyskov I mostly ran this locally on Ubuntu 24.04 on ARM. I also tried it on the host ARM Mac platform, but I'm not 100% sure I waited for an entire LLVM build there.

lyskov (approved):
Looks good to me, - thank you for putting this together @adamnovak !

PR #312, opened 2025-01-31
add support for arbitrary nested namespaces

mattangus:


source/util.cpp
```diff
@@ -36,25 +36,21 @@ namespace binder {
...
+    std::vector<std::string> res;
+
+    while ((pos_end = s.find(delimiter, pos_start)) != std::string::npos) {
+        token = s.substr (pos_start, pos_end - pos_start);
+        pos_start = pos_end + delim_len;
+        res.push_back (token);
+    }
+
+    res.push_back (s.substr (pos_start));
+    return res;
```
lyskov, line 51: Could you please say a few words why do you want to rewrite this function? Thanks,

lyskov (commented):
Thank you for putting this together @mattangus ! Could you please elaborate on why this PR is needed, ie what exactly does not work properly in current version of the code? Maybe add minimal header-file-snippet to PR description that will trigger errors/wrong-behavior?
Thanks,

mattangus:
I wrote this code a while ago but just never got around to making the PR. I did find an edge case but I can't remember what it is now. I've run the new test that I created on `master` and it's producing the right output. I'll find the code that has that edge case and update the test!

PR #313, opened 2025-05-10
Ensure Binder compatibility with LLVM monorepo structure

LachlanGreen:
Included is a update to the llvm path in the final build command. This is necessary because while LLVM 6.0.1 originally used a multi-repo structure, the llvmorg-6.0.1 tag has since been retroactively added to the monorepo. As a result, the build must now target the llvm/ subdirectory inside llvm-project.


lyskov (approved):
Looks good to me, - thank you for fixing this @LachlanGreen !

PR #327, opened 2025-07-26
Refine build tooling

lczech:
This PR adds a bit of helpful functionality for building Binder during dev:

 - `build.py` and `build-and-run-tests.py` have a new flag `--update-binder` which re-compiles Binder without re-compiling LLVM. This is faster for development when using these scripts to build Binder.
 - `test/self-test.py` has a new flag `--accept-all` that accepts all changes, instead of asking about each change. `git diff` can then be used to show all the changes.

Some further small fixes:

 - `test/T42.stl.names.multi.hpp` was missing a header include for `string`.
 - Update the `_pybind11_version_` in `build.py` to the latest commit of the RosettaCommons Pybind11 fork.
 - Add a GitHub Actions test case of Ubuntu using the `build-and-run-tests.py` script.

### DIFF
--- a/build.py
+++ b/build.py
@@ -154,7 +154,7 @@ def install_llvm_tool(name, source_location, prefix_root, debug, compiler, jobs,
     if signature == disk_signature:
         print('LLVM:{} + Binder install is detected at {}, skipping LLVM installation and Binder building procedures...\n'.format(llvm_version, build_dir))
 
-    else:
+    elif not update_binder:
         print('LLVM build detected, but config/binder version has changed, perfoming a clean rebuild...')
         if os.path.isdir(build_dir): shutil.rmtree(build_dir)
 
@@ -285,7 +285,7 @@ def main(args):
     parser.add_argument('--llvm-version', default=None, choices=['6.0.1', '13.0.0', '14.0.5'], help='Manually specify LLVM version to install')
     parser.add_argument('--annotate-includes', action="store_true", help='Annotate includes in generated source files')
     parser.add_argument('--trace', action="store_true", help='Binder will add trace output to to generated source files')
-    parser.add_argument('--update-binder', action="store_true", help='Recompile the Binder binary; useful when working on the Binder code')
+    parser.add_argument('--update-binder', action="store_true", help='Recompile only the Binder binary; useful when working on the Binder code')
 
     global Options
     Options = parser.parse_args()


.github/workflows/main.yml
```diff
@@ -7,6 +7,24 @@ on:
...
+        set -x
+        uname -a
+        cat /etc/issue
+        sudo apt install python3 python3-dev clang make cmake
+    - name: Checkout
+      uses: actions/checkout@v4.0.0
+    - name: Compile and Test
+      run: |
+        python3 build-and-run-tests.py -j 2 --llvm-version 14.0.5
+
```
lyskov, line 27: Could you please use something like `ubuntu-24.04` instead of latest? We do not want this to suddenly start failing if/when GitHub update latest-runner image.

build.py
```diff
@@ -29,8 +29,8 @@
 
 _python_version_ = '{}.{}'.format(sys.version_info.major, sys.version_info.minor)  # should be formatted: 2.7, 3.5, 3.6, ...
 
-_pybind11_version_ = 'aa304c9c7d725ffb9d10af08a3b34cb372307020'
-
+#_pybind11_version_ = 'aa304c9c7d725ffb9d10af08a3b34cb372307020'
+_pybind11_version_ = 'a2e59f0e7065404b44dfe92a28aca47ba1378dc4'
```
lyskov, line 33: What is the reason for this change?

build.py
```diff
@@ -154,7 +154,7 @@ def install_llvm_tool(name, source_location, prefix_root, debug, compiler, jobs,
     if signature == disk_signature:
         print('LLVM:{} + Binder install is detected at {}, skipping LLVM installation and Binder building procedures...\n'.format(llvm_version, build_dir))
 
-    else:
+    elif not update_binder:
```
lyskov, line 157: This seems a bit odd, - if signature does not match then it is clear that Binder upgrade is needed. Could you please elaborate on the logic here? My initial impression for a new `--update-binder` option would be that it will trigger script to check if such upgrade is needed and rebuild Binder if so. Do you see this differently? If so please elaborate.

test/self-test.py
```diff
@@ -151,7 +154,8 @@ def main():
 
     parser.add_argument('--pybind11', default='', help='Path to pybind11 source tree')
 
-    parser.add_argument("--accept", action="store_true", help="Run tests and accept new tests results as reference")
+    parser.add_argument("--accept", action="store_true", help="Run tests and ask to accept new tests results as reference")
```
lyskov, line 157: Since we on this, probably "optional" since it is not directly related to new code: name of this option now clearly misleading, especially with introduction of `--accept-all`. How about something like `--interactive-accept` (I am open to suggestions if you have better ideas).

lyskov (changes requested):
Thank you for PR @lczech ! Looks good to me, please see a few comments/suggestions below. Thanks,

.github/workflows/main.yml
lczech: Sure, will do!

test/self-test.py
lczech: Yep, `--interactive-accept` sounds more descriptive, good idea! Didn't want to change compatibility there, but if you are happy with that, I can change this.

build.py
lyskov: Hm… it sounds like there is a disconnect here, please let me know if below is not what you are observing on your system:
-- there should not be automatic rebuild unless binder repository have a new commits, - ie changing code should not trigger the rebuilds.
-- to avoid re-build checks (and potential rebuilds) the simplest way is to pass `--binder <path-to-binary>` option when using `build-and-run-tests.py`.

Note that I do like the idea of having `--update-binder` option, but I though it will trigger be something like "rebuild binder binaries even if no new commits are found" so its implementation should probably just call `ninja` in the right dir (this is what usually do by hands)

test/self-test.py
lyskov: I am fine with changing compatibility here, so please feel to rename it. Thanks,

Issue #324, opened 2025-07-24
Tests broken under Ubuntu 24 due to holder type


lczech:
Hi @lyskov,

I was trying to start implementing some of my previous suggestions (#319, #320, #322, and maybe #323), and wanted to run the Binder tests as a baseline. However, on Ubuntu 24, I'm getting the following errors:

```
> python3 build-and-run-tests.py -j 8

[...]

T00.basic.hpp Compiling binding results...
cd /home/lucas/Downloads/binder/test/build && clang++ -O3 -shared -std=c++11 -isystem /home/lucas/Downloads/binder/build/pybind11/include -I/usr/include/python3.12 -I./.. -I./../.. -I./../../source T00_basic.cpp -o T00_basic.so -fPIC

T00.basic.hpp Testing imports...
cd /home/lucas/Downloads/binder/test/build && /usr/bin/python3 -c 'import T00_basic'

Comparing results for test T00.basic.hpp...
diff /home/lucas/Downloads/binder/test/T00.basic.ref.cpp /home/lucas/Downloads/binder/test/build/T00_basic.cpp
15,16c15,16
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>)
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*)
---
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>, false)
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*, false)
82,83c82,83
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>)
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*)
---
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>, false)
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*, false)
104,105c104,105
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>)
< 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*)
---
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>, false)
> 	PYBIND11_DECLARE_HOLDER_TYPE(T, T*, false)
126c126
< typedef std::function< pybind11::module & (std::string const &) > ModuleGetter;
---
[...]

PR #328, opened 2025-07-26
Update test references

lczech:
As mentioned in #324, the tests do not seem to be up to date with what Binder generates any more. This PR attempts to fix this. There are some major changes to some of the test cases though - I'd appreciate your feedback @lyskov on whether this is expected and in order. If not, let me know how to fix this :-)


lyskov:
Thank you for PR @lczech ! This is a quite a big difference, - let me run the ~same thing locally on current master and see if i can reproduce this.

lyskov:
> that? Binder is using LLVM, so where would GCC come into play when creating the test references?

-- The code will be always produced by Binder and so with LLVM, however you can (try) compile it with any compiler including GCC (and this is where the error that I mentioned will manifest).

Just to double check: you used the default LLVM version when running `build-and-run-tests.py` script?

lczech:
> The code will be always produced by Binder and so with LLVM, however you can (try) compile it with any compiler including GCC (and this is where the error that I mentioned will manifest).

Ah I see what you mean. Yes, I am also compiling on GCC, and AppleClang, see, e.g., [here](https://github.com/lczech/genesis/actions/runs/16662498744) - all of that works fine except for the int sizes issue #258

> Just to double check: you used the default LLVM version when running build-and-run-tests.py script?

I'm using this:

```
python3 build.py -j 8 --llvm-version 14.0.5 --update-binder
python3 build-and-run-tests.py -j 8 --llvm-version 14.0.5
```

which includes my build tool improvement from #327.

lczech:
Right, that makes sense! I guess #337 does this anyway already, so this PR can be closed then?!

lyskov:
> Right, that makes sense! I guess #337 does this anyway already, so this PR can be closed then?!

probably yes

Issue #319, opened 2025-07-21
Custom return value policy per function


lczech:
Hey Sergey,

if I'm understanding the [documentation](https://cppbinder.readthedocs.io/en/latest/config.html) correctly, the return value policies for each of the categories (`default_...`) are applied to all functions in that category, and there is no way to specify a specific return value policy per function or member?

How would I then override this for functions for which the particular default is not the right policy? I cannot find anything on that.

Thanks and so long
Lucas

PR #329, opened 2025-07-27
Add more return value policy support

lczech:
This PR adds more support for return value policies (RVP):
    
   - Add RVP defaults for free functions. They were mixed in with static functions before.
   - Add RVP default for assignment operators. See below for details.
   - Add RVP customization per function. That's the main addition I was after, see #319
   - Add test case and documentation for all of the above.

The RVPs for assignment operators (`operator=`, but also all the `operator+=` etc, see [here](https://en.cppreference.com/w/cpp/language/operator_assignment.html)) are an interesting case: When they are defaulted (implemented as `= default;`), they implicitly end with `return *this;`, returning the called instance itself. Thus, the pybind RVP `automatic` is not the best choice for them - `reference_internal` is a better fit. Hence this addition, to separate those special functions from others. 

However, to not break backwards compatibility with previous Binder versions, I have implemented this such that `default_member_assignment_operator_return_value_policy` is empty by default and not used unless explicitly set in the Binder config by the user. This is the least disruptive for existing bindings - although I'd argue that introducing a breaking change here and defaulting this to `reference_internal` might be a good choice in terms of "least surprise" for users. It would probably not break any bindings code to go from `copy` to `reference_internal` here, as likely no one in Python will use the return value of the assignment operator anyway... But still it felt like the right thing to do. @lyskov, let me know what you think.

While at it, I also refined some other parts:

   - Add some new headings and refinements to config doc, just to clear it up a bit.
   - Add checks for correct +/- in config token parsing - those were missing for many tokens.

I hope that was okay to mix this in here - @lyskov, let me know if you prefer that in a different PR instead.

Lastly, while playing around with this, I noticed that there are defaults for the return value policies of rvalue references, but Binder never actually emits bindings for such functions... I kept everything as it was (and even added rvalue references for free functions for completeness), but I was wondering if it's worth having those around. Or am I missing something and functions returning rvalue references can indeed be bound by Binder? If so, how?

Cheers and looking forward to hearing your thoughts
Lucas

### DIFF
--- a/documentation/config.rst
+++ b/documentation/config.rst
@@ -1,15 +1,14 @@
 Configuration
 #############
 
-Binder provides two ways to supply configuration options: command-line and config file.
+Binder provides two ways to supply configuration options: command-line options for overall options and a config file for specific details.
 
 
 
 Command line options
 ====================
 
-``--root-module`` specify name of generated Python root module. This name is also used as prefix for various Binder output
-files. Typically the following files will be generated: ``<root-module>.cpp``, ``<root-module>.sources``,
+``--root-module`` specify name of generated Python root module. This name is also used as prefix for various Binder output files. Typically the following files will be generated: ``<root-module>.cpp``, ``<root-module>.sources``,
 ``<root-module>.modules``.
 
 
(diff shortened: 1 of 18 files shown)


lczech:
> My priority on adding this will be to make sure [a] backward compatibility and [b] that defaults should be safe to use even if it cost some performance.

Almost. Except for the static member functions, none of this should break existing Binder configs. Only if any of the new tokens are specified will they be used.

> could you please elaborate on this? Ie in my view static function is effectively a free function that live in different namespace. Right now my vote will be to not add this one.

Well, they are not the same as far as C++ is concerned. Separating them gives a bit more flexibility, and is clearer. It was rather confusing to me to figure out that the RV policy setting for free functions is the "static" one... However, as mentioned, if I recall correctly, it might however not be backwards compatible, so that might need some tweaking if you want to fully maintain backwards compatible...

> re choosing reference_internal default_member_assignment_operator_return_value_policy: I am not sure about this, automatic seems to be a safe hence will be what I want to choose.

> re default_member_assignment_operator_return_value_policy: the more I think about it then more it looks like this is highly depended on case-by-case basis. Could you please provide a justification why this category should have a separate config option?

Assignment operators return an lvalue reference to `*this`, which the `automatic` policy will turn into a copy, which is wrong from a semantic point of view. I am not sure how needed this is, as likely no one will use it that way in Python (e.g., `(a += 1) += 2` or the like), but it seems like the right thing to do in terms of code semantics. I can see though that this might be confusing.

> EDIT: and why just heaving " RVP customization per function" is not enough?

That feature allows to do this as well of course. But assignment operators are so common that I thought it might be worth defaulting all of them at once to a reasonable RV policy.

lyskov:
Want to revisit this, re:
> Assignment operators return an lvalue reference to *this, which the automatic policy will turn into a copy, which is wrong from a semantic point of view.

-- well, what else can Python do besides copy at this point? I mean this seems to be the only safe choice for default, would you agree?

PR #331, opened 2025-07-30
Allow multiple add ons for the same class or namespace

lczech:
Currently, when specifying the `add_on_binder` or `add_on_binder_for_namespace` tokens multiple times, later ones silently replace previous ones, which is unexpected. This PR extends this to capture multiple add ons, which can be useful to break them down into smaller pieces.

### DIFF
--- a/source/class.cpp
+++ b/source/class.cpp
@@ -1407,8 +1407,12 @@ void ClassBinder::bind(Context &context)
 
 	c += bind_repr(context, Config::get());
 
-	std::map<string, string> const &external_add_on_binders = Config::get().add_on_binders();
-	if( external_add_on_binders.count(qualified_name_without_template) ) c += "\n\t{}(cl);\n"_format(external_add_on_binders.at(qualified_name_without_template));
+	std::map<string, std::vector<string>> const &external_add_on_binders = Config::get().add_on_binders();
+	if( external_add_on_binders.count(qualified_name_without_template) ) {
+		for( auto const& add_on : external_add_on_binders.at(qualified_name_without_template) ) {
+			c += "\n\t{}(cl);\n"_format(add_on);
+		}
+	}
 
 	c += bind_nested_classes(context);
 
--- a/source/config.cpp
+++ b/source/config.cpp
@@ -196,7 +196,7 @@ void Config::read(string const &file_name)
 
 			if( bind ) {
 				auto binder_function = split_in_two(name, "Invalid line for add_on_binder specification! Must be: name_of_type + <space or tab> + name_of_binder. Got: " + line);
-				add_on_binders_[binder_function.first] = trim(binder_function.second);
+				add_on_binders_[binder_function.first].push_back(trim(binder_function.second));
 			}
 		}
 		else if( token == _binder_for_namespace_ ) {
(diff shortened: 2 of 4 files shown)


lyskov:
Thank you for PR! Could you please elaborate on how this will be used? Is the main idea is to allow user to add separate functions? If so, what will be he call order of these functions? Also, could you please elaborate on how this will be better to just supplying a single function from which user can call other functions?

lczech:
Hey @lyskov,

> Thank you for PR! Could you please elaborate on how this will be used? Is the main idea is to allow user to add separate functions? 

Exactly that. I am structuring my binder config file in sections, with a small section for each header file of the code I am binding. So for header `abc.hpp` of my library, I'm adding a few lines for all configuration for functions and classes in that header. This keeps things neatly ordered. However, these headers might all be part of the same namespace. With this PR, I can now add extra bindings to that namespace per header.

As for classes: Not sure if this is strictly needed as well, but I figured that while at it, I might as well implement that as well. At the least, silently overwriting previous add ons is unexpected, and this avoids that. I don't currently have a use case for classes, but others might.

> If so, what will be he call order of these functions? 

In the order they are defined in the config. The PR just replaces the string with a vector, and adds all of them in order. Not more than that.

> Also, could you please elaborate on how this will be better to just supplying a single function from which user can call other functions?

That would work as well I suppose, but would then not allow the setup above that allows for clear separation of concerns in the binder config, and instead would add a level of indirection for that function, which might be harder to trace and less flexible.

Hope that clarifies it, and sounds useful :-)
Lucas

lczech:
Not sure that I quite understand what you are asking.

> Ie why defining extra-bindings-function definition somehow got tight-up to project source headers? Or do you place these extra-bindings-functions into project original headers, is that the idea?

My extra code needed for Binder is kept separate from my original project headers, to keep things as cleanly separated as I can. But in order to bring structure into this, I am "mirroring" the directory structure of the original code, and for each header that needs extra binding code also generate a header (and if needed source file) with the same name, which then contains the functions used for the `add_on_binder` call.

For instance [this](https://github.com/lczech/genesis/blob/17adec3d885393d60c294b20d23549e75cc03406/python/binder.cfg#L300) and [this](https://github.com/lczech/genesis/blob/17adec3d885393d60c294b20d23549e75cc03406/python/binder.cfg#L314) line are both using `add_on_binder_for_namespace` for the same namespace. The functions are defined in two files [here](https://github.com/lczech/genesis/tree/python/python/custom/genesis/util/io) (`input_source_add_ons.hpp` and `output_target_add_ons.hpp`) in the directory where all my custom bindings code for that namespace is. So there are two functions that provide add ons for the same namespace, because those add ons are covering two different original headers. Just to keep things separate. Otherwise, I'd have one large add on file for the namespace, which I think could quickly become overwhelming.

Does that make sense? Or is there a cleaner way to do this? Either way, I think the functionality added here is reasonable, as otherwise, add-ons on the same namespace would silently overwrite each other, which is unexpected.

lyskov (changes requested):
Looks good to me, thank you @lczech ! A few things before we could merge this:

-- could you please add documentation for a new config option and update docs for existent one? Please particularly highlight calling order.

-- could you please add minimal test(s) for this? I imagine something simple like single namespace and single class inside it with both having two add-on-binders.

-- run all tests and double check that there is no unexpected changes.

Thanks,

lyskov:
@lczech bumping this PR, please see comments above. Thanks,

